PR #20, opened 2022-12-03
Separation logic support

FrogOfJuly:
# Road map
The idea is to add as much separation logic support to liquid java as is supported by CVC5. I will add items here as my understanding of the path increases. 
## Examples
- [x] Annotations
  - [x] HeapRefinement
- [ ] Annotated interfaces
  - [x] FileWriter
  - [x] Stream
  - [ ] Linked List
  - [ ] Binary Tree
- [ ]  Annotation language
  - [x] points-to 
  - [x] values versus pointer
  - [x] separating conjunction
  - [x]  empty heap
  - [ ] nil

Results:
[Separation_logic_support_in_Liquid_Java_LASIGE.pdf](https://github.com/CatarinaGamboa/liquidjava/files/11708724/Liquid_java_LASIGE.1.pdf)


alcides:
We are going with a different direction for liquid java: using uniqueness types.
But this branch is important as an alternative and should be archived.

PR #29, opened 2023-10-26
Autoformat repository on commit

alcides:
Installed a pre-commit hook.


GUIpsp:
The gitlab runner is probably running low on memory, causing catastrophic garbage collection behaviour

PR #38, opened 2024-12-01
Formatting hooks

GUIpsp:
The current pre-commit hook is next to useless, as no one uses it. In addition, the repository contains a mixture of formats and indentation.

These commits enable a set of more useful hooks, and the palantir formatter. I have not included a commit reformatting everything (it would be sizeable), but any future commit with pre-commit enabled will obey the formatting.


CatarinaGamboa (approved):

alcides:
Actually, it might be a good time to apply that repo-wide formatting diff.

PR #46, opened 2025-07-11
Ghost boolean

rodrigomilisse:
Added basic support for boolean ghost variables. Only tested in trivial case


CatarinaGamboa (changes requested):
This looks good! Just change the variables used in the test to make it more consistent with the variable names.
Also, the test should be in the `testSuite`, and make sure that it is run in CI (you can check if the filename is tested).

CatarinaGamboa:
@rodrigomilisse are you still planning on making changes here before we merge?

PR #52, opened 2025-09-22
Qualify Ghost Names

rcosta358:
### Summary
Added qualified names so ghosts/states with the same name from different classes don't conflict with each other.

### Motivation
Different classes frequently reuse ghost/state names (e.g., `open`, `close`, `size`). Without qualification, simple name lookups collide, causing wrong substitutions and sort mismatches in the SMT solver.

### Key Changes

- **Qualified names:** ghosts/states are stored and looked up with their declaring class prefix (e.g., `java.util.ArrayList.size`, `com.example.SM1.initialized`).
- **Keep built-ins unqualified:** `old`, `length`, `addToIndex` and `getFromIndex`.
- **Safe resolution:** `VCChecker` filters candidate ghosts/states to those whose prefix matches the types (and supertypes) of variables in the current VC, preventing unrelated classes with the same simple name from matching (removes ambiguity).
- **Hierarchy-aware:** added functionality in `AuxStateHandler` so subclasses can reference superclass states (`Bus.close` -> `Car.close`)
- **Simple name fallback:** added fallback in `TranslatorToZ3`to match by simple name and number of parameters, preferring exact qualified names.

**Note:** The last three points were implemented using an AI coding agent (GPT-5).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,12 @@
+{
+  "configurations": [
+    {
+      "type": "java",
+      "name": "Run LiquidJava",
+      "request": "launch",
+      "mainClass": "liquidjava.api.CommandLineLauncher",
+      "projectName": "liquidjava-verifier",
+      "args": "liquidjava-example/src/main/java/test"
+    }
+  ]
+}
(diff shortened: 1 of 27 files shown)


CatarinaGamboa (approved):
This is a hugeee PR. I’m gonna stamp it for now since all tests are passing, it looks good overall and we discussed it offline, but we should try to split them into smaller pieces. For example, 1)add the new methods needed; 2) change call sites and add tests 3) remove unused code

PR #54, opened 2025-09-24
Analyze All Packages In Directory

rcosta358:
Previously the verifier only processed the last package in the directory given, skipping the others and leading to false positives. Now it analyzes every package in the directory.
However, this could be an issue when dealing with large codebases, making the verification very slow.


CatarinaGamboa (approved):
Lgtm. Discussed over chat.

PR #76, opened 2025-11-04
Check Method Signatures in External Refinements

rcosta358:
This is an improvement to #66. In external refinements, instead of just checking the name and parameter count of methods, we now check their whole signatures, including name, parameter count, parameter types and return type.

When a method cannot be found with the exact signature, a list of methods with the same name is provided in the error message, if available.

**Example:**
```
Found Error: Could not find method 'int acquire()' for 'java.util.concurrent.Semaphore'
Available overloads:
  void acquire()
  void acquire(int)
```


rcosta358:
One test is failing because the workflow tests the PR against a merged version with main, which includes that test that only passes with the changes in main.

CatarinaGamboa (approved):
Great! This is very helpful and the error message is awesome 🎉

PR #77, opened 2025-11-04
Fix Refined Method Overloading

rcosta358:
There was still an issue regarding refinements in overloaded methods, where the typechecker mistakenly looked up the wrong method. `RefinedFunction` now stores the fully qualified method signature and uses it in equality/hash logic in order to distinguish between overloaded methods in the context. Also added a test that previously passed and now successfully fails.


CatarinaGamboa (approved):
nicee

PR #81, opened 2025-11-05
Add Diagnostics Classes

rcosta358:
This PR adds a class for each type of error, instead of mapping all errors to strings. Warnings were also introduced.  This change will make diagnostics more structured, extensible and easier to manage, which the VS Code extension will benefit from.

### Changes
- Each error and warning has its own class with their respective information and `toString` implementation
  - *`LJError`*
    - `RefinementError` - when a refinement is violated
    - `StateRefinementError` - when a state refinement is violated
    - `NotFoundError` - when an element is not found
    - `SyntaxError` - when the syntax is invalid
    - `GhostInvocationError` - when a ghost call as wrong arguments
    - `StateConflictError` - when states are disjoint
    - `IllegalConstructorTransitionError` - when state refinement in constructor has `from` transition
    - `InvalidRefinementError` - when refinement is semantically invalid (e.g., not a boolean expression)
    - `CustomError` - any other error
  - *`LJWarning`*
    - `ExternalClassNotFoundWarning` - when class in external refinement is not found
    - `ExternalMethodNotFoundWarning` - when method in class of external refinement is not found
- Renamed the `errors` folder to `diagnostics`

### Next Steps
- Integrate these new diagnostic classes throughout the codebase
- `LJDiagnostics` will replace the `ErrorEmitter`
- `ErrorHandler` will no longer be needed

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/CustomError.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/CustomError.java
@@ -2,6 +2,7 @@ package liquidjava.diagnostics.errors;
 
 /**
  * Custom error with an arbitrary message
+ * 
  * @see LJError
  */
 public class CustomError extends LJError {
--- a/liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/GhostInvocationError.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/GhostInvocationError.java
@@ -5,6 +5,7 @@ import spoon.reflect.declaration.CtElement;
 
 /**
  * Error indicating that a ghost method invocation is invalid (e.g., has wrong arguments)
+ * 
  * @see LJError
  */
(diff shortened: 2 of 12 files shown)


liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/LJError.java
```diff
@@ -0,0 +1,65 @@
+package liquidjava.diagnostics.errors;
+
+import liquidjava.diagnostics.ErrorPosition;
+import liquidjava.utils.Utils;
+import spoon.reflect.cu.SourcePosition;
+import spoon.reflect.declaration.CtElement;
+
+// base class for all LiquidJava errors
```
CatarinaGamboa: you can have this comment as proper javadoc

CatarinaGamboa (approved):
LGTM and the next steps seem great - thanks for splitting the change in multiple PRs!
Lets just add some javadoc to the multiple types of errors you created so everyone can understand when to use each type of error

PR #89, opened 2025-11-06
Skip String Literals in OperationsChecker

rcosta358:
This PR is similar to #63.
There was a problem where this simple program failed parsing:

```java
public class Example {
    void example() {
        System.out.println("Hello, " + "World!");
    }
}
```

```
Syntax error with message
Found 1 error, with the message:
* Error in no viable alternative at input 'Hello,', in the position 5
```

This was because strings were not being handled in the binary operations, in this case string concatenation.
It was fixed by skipping string literals in the `OperationsChecker`.


CatarinaGamboa (approved):
Good catch! What a strange error 😅

Issue #105, opened 2025-11-08
Include error message in test suite checks


CatarinaGamboa:
Our current tests only check if the verification fails or not. However, it is also important to see if it fails with the intended error message, to check if the error is the expected one and not an alternate one.

PR #107, opened 2025-11-08
feat(tests): Check for specific error messages in test suite

subha0319:
Implements #105.

Updates the `TestExamples` test suite to check for specific error messages.
The test harness now reads a comment tag `// @ExpectedError: "Error Title"` from files that are expected to fail and validates the actual error title against the expected one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Additionally, you'll need the following dependency, which includes the LiquidJav
 #### Maven
 ```xml
 <dependency>
-    <groupId>io.github.rcosta358</groupId>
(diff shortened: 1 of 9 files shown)


subha0319:
Hi @CatarinaGamboa,  Thanks for the feedback.

I've added the `ExpectedError` annotation to a couple more test files as you suggested: `ErrorAfterIf.java` and `ErrorArithmeticBinaryOperations.java` (along with `ErrorSimpleAssignment.java` which I had updated earlier).

And I've made sure to place the annotations on line 1 in all the examples.

Let me know if there's anything else.

PR #109, opened 2025-11-09
Code Refactoring

rcosta358:
Cleaned up some code by:
- Moving scattered constants into dedicated classes: `Formats`, `Keys`, `Ops`, `Patterns` and `Types`
- Refactoring some if else statements to switch cases instead
- Updating some switch cases to use arrow syntax instead
- Commenting out some unused methods (to be potentially removed in the future)

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/StateConflictError.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/StateConflictError.java
@@ -14,7 +14,9 @@ public class StateConflictError extends LJError {
     private String className;
 
     public StateConflictError(CtElement element, Predicate predicate, String className) {
-        super("State Conflict Error", "Found multiple disjoint states in state transition", element);
+        super("State Conflict Error",
+                "Found multiple disjoint states in state transition — State transition can only go to one state of each state set",
+                element);
         this.predicate = predicate;
         this.className = className;
     }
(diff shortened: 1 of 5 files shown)


liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/StateConflictError.java
```diff
@@ -14,7 +14,7 @@ public class StateConflictError extends LJError {
     private String className;
 
     public StateConflictError(CtElement element, Predicate predicate, String className) {
-        super("State Conflict Error", "Found multiple disjoint states from a StateSet in refinement", element);
+        super("State Conflict Error", "Found multiple disjoint states in state transition", element);
```
CatarinaGamboa: We could even improve this message to something like "Found multiple disjoint states in state transition - State transition can only go to one state of each state set"

liquidjava-verifier/src/main/java/liquidjava/rj_language/Predicate.java
```diff
@@ -221,36 +223,29 @@ public ValDerivationNode simplify() {
...
-        else if (type.equals(Utils.LONG))
-            ex = new LiteralReal(value);
-        else // if(type.equals(Utils.DOUBLE))
-            ex = new LiteralReal(value);
-        return new Predicate(ex);
+        Expression exp = switch (type) {
+        case Types.BOOLEAN -> new LiteralBoolean(value);
+        case Types.INT, Types.SHORT -> new LiteralInt(value);
+        case Types.DOUBLE, Types.LONG -> new LiteralReal(value);
+        default -> new LiteralReal(value);
```
CatarinaGamboa: in the future if we want to add strings this `default` can be a problem, maybe we can emit an error or smt

CatarinaGamboa (changes requested):
looks good, check the comments

CatarinaGamboa (approved):
In a followup for cleanup feel free to remove all code in comments, its just polluting the codebase

PR #110, opened 2025-11-10
Replace `ErrorEmitter` with `LJDiagnostics`

rcosta358:
This PR is a follow-up of #81 and based on #109 (already performed rebase).

- The `ErrorHandler` was removed
- The `ErrorEmitter` was replaced by `LJDiagnostics`, which is now referenced statically to avoid passing it as an argument throughout the code

Errors still need to be refactored for better clarity.


CatarinaGamboa (approved):
left some comments but nothing major, feel free to address them or leave them for another time

PR #117, opened 2025-11-13
Prevent Index Out-Of-Bounds in Function Call With Empty Args

rcosta358:
This PR is related to #86. Instead of adding `this` as default argument when none are provided, we throw a `ParsingException` with the appropriate message instead of failing with index out-of-bounds later on.

Before:
```
Index 0 out of bounds for length 0
```

After:
```
Syntax error with message
Function call cannot have empty arguments
Found in refinement:
size() > 0
In:
@liquidjava.specification.StateRefinement(from = "size() > 0")
void testMethod() {
}
```


CatarinaGamboa (changes requested):
small change

PR #119, opened 2025-11-13
Refactor Diagnostics

rcosta358:
This PR is a follow-up of #110.
It refactors errors and warnings and also gives them a better look.

image

image

### Additional Improvements
- Use annotation position when syntax error occurs, instead of the target
- Exclude qualified names from error messages
- Remove `true && ...` from found ghost states

### Next Steps
Allow for multiple errors by instead of stopping verification when an error is found, throw the error and catch it on visitors (`visitCtMethod`, `visitCtClass`, `visitCtField`), add it to the collection and keep the verification going.


CatarinaGamboa (approved):
great!

PR #123, opened 2025-11-20
Error Refactoring

rcosta358:
This PR simplifies the error handling in the verifier for better clarity and also aids the use of the diagnostic classes in the extension.
 
These changes include:
- Reusing the same logic for throwing errors in `VCChecker`, which was being handled in two different ways
- Using `SourcePosition` directly instead of `CtElement`
- Replacing `details` field with `getDetails` method in `LJDiagnostic` which each diagnostic class can override to show extra error info
- Remove `GhostInvocationError` and use `CustomError` instead

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/diagnostics/LJDiagnostic.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/diagnostics/LJDiagnostic.java
@@ -31,7 +31,7 @@ public class LJDiagnostic extends RuntimeException {
     }
 
     public String getDetails() {
-        return "";
+        return ""; // to be overridden by subclasses
     }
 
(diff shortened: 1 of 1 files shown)


liquidjava-verifier/src/main/java/liquidjava/diagnostics/LJDiagnostic.java
```diff
@@ -59,7 +61,8 @@ public String toString() {
         }
 
         // details
-        if (details != null && !details.isEmpty()) {
+        String details = getDetails();
+        if (!details.isEmpty()) {
```
CatarinaGamboa: if we remove the function this is not needed

CatarinaGamboa (approved):

PR #125, opened 2025-11-21
Simplification Fixes

rcosta358:
This PR improves the simplification of the predicates.
 
- The simplification is now applied until a **fixed point** is reached (no further simplifications are possible) while preserving previous simplifications
- The variable resolver now only extracts variables used in the expression (excluding their own definitions)
- The expression simplifier is now also able to collapse duplicate conjunctions into a single one (`x && x` → `x`) as well as symmetrical ones (`x == y && y == x` → `x == y`)
- Simplify not only `found` but also `expected` predicate in refinement errors

Also added tests such as:
- `x == 1`  →  `x == 1` (does not simplify)
- `x == 1 && x == 1`  →  `x == 1` (collapse duplicate)
- `x == y && y == x`  →  `x == y` (collapse symmetrical)
- `x == 1 && y == 2` →  `x == 1 && y == 2` (does not simplify)
- `x == -y && y == a / b && a == 6 && b == 3` →  `x == -y && y == 2`  →  `x == -2` (two simplifications until fixed point)

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
@@ -52,8 +52,8 @@ public class VCChecker {
         }
         boolean isSubtype = smtChecks(expected, premises, element.getPosition());
         if (!isSubtype)
-            throw new RefinementError(element.getPosition(), expectedType.getExpression(),
-                    premisesBeforeChange.simplify(), map);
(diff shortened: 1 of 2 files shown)


liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/RefinementError.java
```diff
@@ -12,19 +12,18 @@
  */
 public class RefinementError extends LJError {
 
-    private final String expected;
+    private final ValDerivationNode expected;
```
CatarinaGamboa: nicee

PR #137, opened 2025-12-21
Flatten Variable Derivation Chains

rcosta358:
This PR improves the simplification for expressions by flattening variable derivation chains to avoid redundancy.
When a variable's origin is the same variable instance (e.g., `#y_3` => `#y_2`), we collapse the chain to use the final node directly, eliminating redundant nested `VarDerivationNodes` in the derivation tree.

### Example
```
#x_4 == (-#y_3) * (2) && #y_3 == #y_2 && #y_2 == 3
```

#### Before
```json
...
{
    "value": 3,
    "origin": {
        "var": "#y_3",
        "origin": {
            "var": "#y_2"
        }
    }
}
...
```

#### After
```json
...
{
    "value": 3,
    "origin": {
        "var": "#y_2"
    }
}
...
```

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/utils/Utils.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/utils/Utils.java
@@ -44,8 +44,15 @@ public class Utils {
     }
 
     public static boolean isSameVariable(String var1, String var2) {
-        String name1 = var1.substring(0, var1.lastIndexOf('_'));
-        String name2 = var2.substring(0, var2.lastIndexOf('_'));
+        int index1 = var1.lastIndexOf('_');
+        int index2 = var2.lastIndexOf('_');
+        // no format
+        if (index1 < 0 || index2 < 0) {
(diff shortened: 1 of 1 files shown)


CatarinaGamboa:
I'm just not super sure about this because it will have more disconnect from code right?
Lets say we have:
`#x_4 == (-#z_3) * (2) && #z_3 == #y_2 && #y_2 == 3`

That expression probably came from code like:
```java
int y = 3;
int z = y;
@Refinement(...)
int x = -z *2;
```
If we are say that `#x == (-#y) * (2) && #y_2 == 3` it can leave the user wondering why is `x==y` and we said it was `x==z`. It creates more disconnect between the error message and the code.

Now, if you are mentioning flattening variable instances for the same variable like `y_1` and `y_2`, that makes more sense, since they are only internal representations but if we have a connection through a different variable, I'm not sure we should flatten it

rcosta358:
You're right. Let me try only flattening variable instances for the same variable.

PR #139, opened 2026-01-30
Add support for Long operations inside predicates

CatarinaGamboa:
Previously, we only tested with a small set of long operations, which did not include operations with Long values inside predicates or long values with L.
Example that would not be properly checked:
```java
void testDiv(){
        // we have a mask and want to see if the value in position [-4] is a 2
        @Refinement("((v/4096) % 16) == 2")
        long v = 0x01000000666663666L;
}
```

Now we can check this and see that the value is 3 and not 2 so an error.

This PR makes changes in:
 - New `LiteralLong` AST node (we had LiteralInt, LiteralDouble, etc, no Long)
 - Added `LiteralLong` through the full pipeline: visitor interface, Z3 translation (using mkInt for exact integer         
  semantics), type inference, etc
 - In SMT solver translation to `long`
 - Tests in testsuite for these additions

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/visitors/CreateASTVisitor.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/visitors/CreateASTVisitor.java
@@ -199,10 +199,11 @@ public class CreateASTVisitor {
             return new LiteralString(literalContext.STRING().getText());
         else if (literalContext.INT() != null) {
             String text = literalContext.INT().getText();
-            try {
-                return new LiteralInt(text);
-            } catch (NumberFormatException e) {
-                return new LiteralLong(text);
+            long value = Long.parseLong(text);
+            if (value <= Integer.MAX_VALUE && value >= Integer.MIN_VALUE) {
+                return new LiteralInt((int) value);
+            } else {
+                return new LiteralLong(value);
             }
         } else if (literalContext.REAL() != null)
             return new LiteralReal(literalContext.REAL().getText());


liquidjava-verifier/src/main/java/liquidjava/rj_language/visitors/CreateASTVisitor.java
```diff
@@ -196,9 +197,14 @@ private Expression literalCreate(LiteralContext literalContext) throws LJError {
...
             return new LiteralString(literalContext.STRING().getText());
-        else if (literalContext.INT() != null)
-            return new LiteralInt(literalContext.INT().getText());
-        else if (literalContext.REAL() != null)
+        else if (literalContext.INT() != null) {
+            String text = literalContext.INT().getText();
+            try {
+                return new LiteralInt(text);
+            } catch (NumberFormatException e) {
+                return new LiteralLong(text);
```
rcosta358: So we only convert a long to a `LiteralLong` if it overflows 32 bits, even if it was declared as a long, right?

Also, instead of using a try-catch here, we could do this:
```java
String text = literalContext.INT().getText();
long value = Long.parseLong(text);
if (value >= Integer.MIN_VALUE && value <= Integer.MAX_VALUE) {
    return new LiteralInt(text);
} else {
    return new LiteralLong(text);
}
```
Both are fine for me. Which one do you prefer?

CatarinaGamboa: to have `literalContext.LONG()` i needed to change the grammar in `.g4` and i didn't want to 😅 ‍
We could not catch the exception its true, it probably is more readable so I'll change it! Athough we usually have more examples with ints but its fine

PR #140, opened 2026-02-05
Add Custom Error Messages

rcosta358:
This PR adds support for custom messages in refinements and state refinements, through an extra parameter `msg`:

```java
@Refinement(value="...", msg="...")
@StateRefinement(from="...", to="...", msg="...")
```

### Examples

image

image

image

In the last example, the `close` method is annotated with two state refinements, each with its own message, so both are shown. However, to simplify the error message, users should specify a message in only one of them, e.g., `"The buffered reader should be open or marked to be closed"`.

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/diagnostics/LJDiagnostic.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/diagnostics/LJDiagnostic.java
@@ -14,6 +14,7 @@ public class LJDiagnostic extends RuntimeException {
     private final String customMessage;
     private String file;
     private ErrorPosition position;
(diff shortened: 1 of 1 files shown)


rcosta358:
Maybe like so?

image

liquidjava-api/src/main/java/liquidjava/specification/Refinement.java
```diff
@@ -16,4 +16,6 @@
 public @interface Refinement {
 
     public String value();
+
+    public String msg() default "";
```
CatarinaGamboa: We should add documentation to all these fields in a follow up pr

PR #142, opened 2026-02-06
Add Ghost Dot Syntax

rcosta358:
This PR adds support for dot syntax in ghost variables and states while keeping it backward compatible with the previous syntax. For this, a new non-terminal `dotCall` was added to the grammar and implemented in the AST visitor.

Currently, all of these are equivalent: `size(this)`, `this.size()`, and `size()`. To refer to the previous value of `size`, we can either use `size(old(this))` or `old(this).size`.

In the future, we plan to drop the functional syntax and only keep the dot syntax to avoid ambiguity and confusion.

### Example

```java
@Ghost("int n")
public class CorrectDotNotationIncrementOnce {

    // explicit this
    @StateRefinement(to="this.n() == 0")
    public CorrectDotNotationIncrementOnce() {}

    // implicit this
    @StateRefinement(from="n() == 0", to="n() == old(this).n() + 1")
    public void incrementOnce() {}
}
```

### DIFF
--- a/liquidjava-example/src/main/java/testSuite/ErrorDotNotationMultiple.java
+++ b/liquidjava-example/src/main/java/testSuite/ErrorDotNotationMultiple.java
@@ -1,0 +1,18 @@
+// Syntax Error
+package testSuite;
+
+import liquidjava.specification.Ghost;
+import liquidjava.specification.Refinement;
+import liquidjava.specification.StateRefinement;
+
+@Ghost("int size")
+public class ErrorDotNotationMultiple {
+
+    @StateRefinement(to="size() == 0")
+    public ErrorDotNotationMultiple() {}
+
+    void test() {
+        @Refinement("_ == this.not.size()")
+        int x = 0;
+    }
+}


liquidjava-verifier/src/main/java/liquidjava/rj_language/visitors/CreateASTVisitor.java
```diff
@@ -171,15 +171,48 @@ private Expression functionCallCreate(FunctionCallContext rc) throws LJError {
...
             if (args.isEmpty())
                 throw new ArgumentMismatchError("Alias call cannot have empty arguments");
+
             return new AliasInvocation(ref, args);
+        } else {
+            return dotCallCreate(rc.dotCall());
+        }
+    }
+
+    private Expression dotCallCreate(DotCallContext rc) throws LJError {
```
CatarinaGamboa: We def need documentation here explaining what the function does

liquidjava-verifier/src/main/java/liquidjava/rj_language/visitors/CreateASTVisitor.java
```diff
@@ -171,15 +171,48 @@ private Expression functionCallCreate(FunctionCallContext rc) throws LJError {
...
+        if (rc.OBJECT_TYPE() != null) {
+            // this.func(args)
+            String text = rc.OBJECT_TYPE().getText();
+            int dot = text.indexOf('.');
+            String target = text.substring(0, dot);
+            String simpleName = text.substring(dot + 1);
+            String name = Utils.qualifyName(prefix, simpleName);
+            List<Expression> args = getArgs(rc.args(0));
+            if (!args.isEmpty() && args.get(0)instanceof Var v && v.getName().equals(Keys.THIS)
```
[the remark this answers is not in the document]
rcosta358: I also don't like it, blame the formatter 😂

liquidjava-verifier/src/main/java/liquidjava/rj_language/visitors/CreateASTVisitor.java
rcosta358: Should I add documentation for all other functions in this file then?

CatarinaGamboa: probably a good idea yeah

CatarinaGamboa (approved):
niceee

PR #145, opened 2026-02-09
Add `#result` for Return Refinements

### LINKED ISSUE #133: Change `return` to `$result` for method return values in refinements
`return` is a statement, and it's strange to have it as a variable to represent the return value.
Therefore, we could change it for `result`. But since result can be an actual variable that represents something other than the return value we, we can try using `$result`.

This would mean a small change in the grammar, probably, and a change in replacing the return value.
Ps: the _ should still work in any case

Example Before:
```java
@Refinement("return > 0") 
int test2 (){
    return 10;
}
```
Example After:
```java
@Refinement("$result > 0") 
int test2 (){
    return 10;
}
```

rajshivu:
Updated CreateASTVisitor to map refinement variables '_' and 'return' to '$result' to prevent conflicts. 
Verified by manual test in ReturnRefinementTest.java.
firstos


rcosta358 (commented):
Hi @rajshivu,
This is not really what we were looking for. Your approach replaces the variables `return` and `_` with `$result`, which is incorrect and causes the tests to fail.

To allow `$result` to represent return values of methods, you need to modify the grammar to allow identifiers to start with `$`, substitute the variable `$result` with `_` (similarly to `return`), and finally add some tests to check if your implementation is correct.

rcosta358 (commented):
Also, if you have any suggestions other than `$result`, which is not very idiomatic for Java, feel free to let us know!

rcosta358 (changes requested):
Hey @rajshivu,
Almost there. Your implementation is not quite correct yet. Currently what you're doing is replacing all `return` and `$result` variables with `_` in the `CreateASTVisitor`. That way, the following code passes, when it shouldn't:

```java
void test() {
  @Refinement("$result > 0") // $result is replaced with `_`, which refers to `x`, so it passes
  int x = 1;
}
```

Actually, `$result` or `return` should only be allowed on refinements that target methods, such as:

```java
@Refinement("$result > 0")
int test() {
  return 1;
}
```

So, we don't actually want to modify the `CreateASTVisitor`. Instead, we want to modify the `MethodsFunctionsChecker`. More specifically, the `handleFunctionRefinements`, and add an additional case for `$result`, similarly to `return`:

```java
private Predicate handleFunctionRefinements(RefinedFunction f, CtElement method, List<CtParameter<?>> params) throws LJError {
  // ...
  ret = ret.substituteVariable("return", Keys.WILDCARD);
  ret = ret.substituteVariable("$result", Keys.WILDCARD); // ADD THIS
  f.setRefReturn(ret);
  // ...
}
```

Finally, please add tests to `liquidjava-example/src/main/java/testSuite`, not `liquidjava-example/src/main/java/testMultiple`. For tests that should pass, they must include "Correct" in their file name, whereas for tests that should fail, they should include "Error", along with the expected error in the first line of the file as a comment. To run the tests locally, you can run `mvn test`.

liquidjava-verifier/src/main/antlr4/rj/grammar/RJ.g4
```diff
@@ -92,7 +92,7 @@ BOOL    : 'true' | 'false';
 ID_UPPER: ([A-Z][a-zA-Z0-9]*);
 OBJECT_TYPE:
 		  (([a-zA-Z][a-zA-Z0-9]+) ('.' [a-zA-Z][a-zA-Z0-9]*)+);
-ID     	: '#'*[a-zA-Z_][a-zA-Z0-9_#]*;
+ID     	: '#'*('$')? [a-zA-Z_] [a-zA-Z0-9_#]* ;
```
CatarinaGamboa: This is not a great idea to change all the IDs to allow the `$`. We can improve it by only allowing $result and only in the places it can appear (not when a ghost or state is created for example)

liquidjava-verifier/src/main/antlr4/rj/grammar/RJ.g4
CatarinaGamboa: we can just treat it as another literalExpression #result that has the specific regex `$result`

CatarinaGamboa (changes requested):
Changing all IDs to allow for the $ will make it harder to pinpoint errors in other variables. Since the only thing we want is to allow $result as a possible var in the refinements we can add it to the place i pointed in the comments. Try the change and test it to see if it works

CatarinaGamboa (changes requested):
Looks better! Merge main into this branch so there are no conflicts and then we fix the nitpick below. Then request a new review and maybe we'll merge it

liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/general_checkers/MethodsFunctionsChecker.java
```diff
@@ -154,16 +154,15 @@ private Predicate handleFunctionRefinements(RefinedFunction f, CtElement method,
...
             RefinedVariable v = rtc.getContext().addVarToContext(param.getSimpleName(), param.getType(), c, param);
-            rtc.getMessageFromAnnotation(param).ifPresent(v::setMessage);
             if (v instanceof Variable)
                 f.addArgRefinements((Variable) v);
             joint = Predicate.createConjunction(joint, c);
         }
         Optional<Predicate> oret = rtc.getRefinementFromAnnotation(method);
         Predicate ret = oret.orElse(new Predicate());
         ret = ret.substituteVariable("return", Keys.WILDCARD);
+        ret = ret.substituteVariable("$result", Keys.WILDCARD);// added for refinement
```
rcosta358: Unnecessary comment.

rcosta358 (changes requested):
There's a minor conflict in `CreateASTVisitor` that does not allow the tests to run.
Also added some small requests.

rcosta358 (approved):

PR #146, opened 2026-02-10
Add Counterexamples to Refinement Errors

rcosta358:
This PR adds counterexamples to refinement errors by extracting the SMT model from Z3.

### Changes
- Added the `getCounterexample` method to `TranslatorToZ3` that extracts variable assignments (for constants only)
- In the `SMTEvaluator`, when the SMT returns `SAT` we get the SMT model and throw the `TypeCheckError` containing the counterexample
- In the caller methods, instead of returning `boolean` it now returns `Counterexample` - if counterexample is null, then the check succeeded, otherwise it failed
- In the `RefinementError`, we override the `getDetails` method to add the counterexamples as extra information in the error message. Also, it filters out fresh variables which don't even exist in the source code, and if the counterexample is the same as the found type (already included in the message), then the counterexample is not shown.
- In `TranslatorToZ3`, we keep track of the function applications and expressions sent to Z3 so we can reconstruct them from the counterexample when dealing with uninterpreted ghost functions

### Examples

image

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/TypeChecker.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/TypeChecker.java
@@ -16,7 +16,6 @@ import liquidjava.processor.facade.AliasDTO;
 import liquidjava.processor.facade.GhostDTO;
 import liquidjava.rj_language.Predicate;
 import liquidjava.rj_language.parsing.RefinementsParser;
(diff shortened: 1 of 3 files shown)


CatarinaGamboa (changes requested):
Awesome! I think we are almost there. I've just left a comment on changing from `boolean` -> `Counterexample` for the return value, and returning `null`'s. This isn't the best practice, see if you understand the comment and can make that change

PR #147, opened 2026-02-11
Custom Messages Fix

rcosta358:
This PR fixes an issue where if the user used an old version of the LiquidJava API without the `msg` param in `@Refinement`annotations, an exception would be thrown: `java.lang.NoSuchMethodException: liquidjava.specification.Refinement.msg()`.


CatarinaGamboa (approved):
lgtm

PR #152, opened 2026-02-14
Skip Duplicate Tests in CI

rcosta358:
This PR updates the Maven workflow to skip the tests during the Maven build in CI, preventing tests from running twice and reducing the build time.


CatarinaGamboa (approved):
love 1 line prs <3

PR #155, opened 2026-02-19
Context History Tracking

rcosta358:
This PR adds context history tracking for the autocomplete, context debugger, and hover information of the VS Code extension. It works by taking a snapshot of the variables in scope before exiting the context (in methods, constructors, ifs and elses), and then everything else (global vars, instance vars, ghosts, states and aliases).


CatarinaGamboa (approved):
Great! Minor nits

PR #156, opened 2026-02-22
Include subpackage checking

CatarinaGamboa:
Previously we were not checking for subpackages for refinements.
Somehow it works in the extension (@rcosta358 maybe you did some trick?) but using the verifier alone we were erasing the information from the parent package refinements and not checking the insides.
For example:
```
| FileRefinements.java
| TestIncorrect.java // would be caught as an error
| test
 -| TestIncorrect2.java // would not be checked with the info from FileRefinements
```
This PR fixes that


rcosta358 (approved):
lgtm. also tested in the extension.

Issue #64, opened 2025-10-21
Method Call Refinements Not Working Correctly With External Refinements


rodrigomilisse:
Currently, refining the return type of the methods via External Refinements does not propagate the refinement to method calls. This causes error when the caller relies on the refined return type to prove the correctness of their code

### Example

```java
import liquidjava.specification.*;

@ExternalRefinementsFor("java.util.concurrent.Semaphore")
public interface SemaphoreRefinements {

@Refinement("_ == true")
	public abstract boolean isFair();

	public abstract void acquire(@Refinement("_ >= 0") int permits) throws InterruptedException;
}
```

```java
import liquidjava.specification.*;

class TestCase {
	public static void expectFairAndConstructFair() {
		Semaphore sem = new Semaphore(1, true);
		
		@Refinement("_ == true")
		boolean mustBeFair = sem.isFair();
		// sem.acquire(-2); // Uncomment to get an error and ensrure the refinements are being processed
	}
}
```

Yields the following error

```
@liquidjava.specification.Refinement("_ == true")
boolean mustBeFair = sem.isFair()

Type expected:(#mustBeFair_1 == true)
Refinement found:true && true
```

The LiquidJava verifier is failing to recognize that the assignment is correct even though the refinements guarantees the return type.

PR #159, opened 2026-02-26
Context History Improvements

rcosta358:
This PR makes some changes to the context history to improve the [autocomplete](https://github.com/liquid-java/vscode-liquidjava/pull/63) and the [context debugger](https://github.com/liquid-java/vscode-liquidjava/pull/64) in the [extension](https://github.com/liquid-java/vscode-liquidjava) by:
- Add the `file` field to ghosts to filter by file
- Add the `annotationPosition` field to `PlacementInCode` to suggest the variable we are refining in the autocomplete

### Other Changes
- Remove the `ErrorPosition` class which is unnecessary since we can use Spoon's `SourcePosition` in diagnostics instead
- In `AuxStateHandler`, always add failing instances and substitute `this` with instance variable

### DIFF
--- a/liquidjava-verifier/pom.xml
+++ b/liquidjava-verifier/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>io.github.liquid-java</groupId>
 	<artifactId>liquidjava-verifier</artifactId>
-	<version>0.0.15</version>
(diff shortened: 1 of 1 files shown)


CatarinaGamboa (approved):
lgtm! discussed offline

PR #166, opened 2026-02-28
Improve Syntax Errors

rcosta358:
This PR improves the syntax error messages of refinements, ghosts, and aliases, and includes example usages of those.

### Examples

image

image

image

In the future we'll include these examples as hints rather than in the error message itself.

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/parsing/RefinementsParser.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/parsing/RefinementsParser.java
@@ -56,7 +56,8 @@ public class RefinementsParser {
         AliasVisitor av = new AliasVisitor(input);
         AliasDTO alias = av.getAlias(rc);
         if (alias == null)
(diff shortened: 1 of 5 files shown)


liquidjava-verifier/src/main/java/liquidjava/rj_language/parsing/RefinementsParser.java
```diff
@@ -35,16 +35,12 @@ public static GhostDTO getGhostDeclaration(String s) throws LJError {
         ParseTree rc = compile(s);
         GhostDTO g = GhostVisitor.getGhostDecl(rc);
         if (g == null)
-            throw new SyntaxError("Ghost declarations should be in format <type> <name> (<parameters>)", s);
+            throw new SyntaxError("Invalid ghost declaration, e.g. \"int size\"", s);
```
CatarinaGamboa: Lets include the whole alias reference like `@Ghost("int size")` to ensure that it is perceptible

CatarinaGamboa (approved):
lgtm

PR #168, opened 2026-03-03
Added enum types and verification

### LINKED ISSUE #23: Add Enums to refinement language and verification as parameters.
## Why?
Enums can be received as parameters and the state changes can depend on the values gotten from the enum parameter.

## Example
From an example from modeling flight controllers code with the library mavsdk ([gitrepo](https://github.com/CatarinaGamboa/flight-controllers-liquid-state))
```java
public enum Mode {
	Photo,
	Video,
	Unkown
}

class Camera {
	public void setMode(Mode mode) {...}
	
	public void takePhoto() {...}
}
```
In this case, the method `setMode` defines the current mode of the camera. If the value in the parameter is `Mode.Photo` the transition of the method should be to photoMode in which the method `takePhoto()` should be allowed. However, if the `Mode `used is a different one, `takePhoto()`  should not be allowed.

## Intended behavior
The idea here is to add enums to the refinements language and their verification as parameters.
Ideally, the piece of code from above should be modeled in the following way:

```java
@StateSet({"photoMode", "videoMode", "noMode"})
class Camera {
	
	@StateRefinement(to="noMode(this)")
	public Camera() {}
	
	
	@StateRefinement(from="noMode(this) && mode == Mode.Photo", to="photoMode(this)")
	@StateRefinement(from="noMode(this) && mode == Mode.Video", to="videoMode(this)")
	public void setMode(Mode mode) {...}
	
	@StateRefinement(from="photoMode(this)")
	public void takePhoto() {...}
	
	
	public static void main(String[] args) {
		// Correct
		Camera st = new Camera();
		st.setMode(Mode.Photo);
		st.takePhoto();

		// Incorrect
		Camera st = new Camera();
		st.setMode(Mode.Video);
		st.takePhoto(); //error
	}
}
```
This would represent a symbolic automata with the following view:
camera1

The language of the refinements to add the enums should be discussed prior to implementation since there could be several options. One possibility would be the one presented above `mode == Mode.Video`, but others could be `value(mode) == Mode.Video` or `video(mode)`.

## High-level steps to change the current behavior
1. Change the refinements grammar to allow the invocation of the enumerate.
2. Change the traversing of the AST to analyze enum definitions and store the values in context
3. Change the traversing of the AST to account for the enumerate as a parameter and store its information in the context.
4. See how the enums information can be translated into the SMT and check if the verification condition works.

cheestree:
Closes #23. This pull request introduces support for Java enums in the refinement type system and SMT translation. The changes include updates to the grammar, AST, visitors, and Z3 translation logic, as well as new test cases for correct and incorrect enum usages.

image

An `enumerate` rule and `ENUM` token for expressions like `Color.Red` and the respective AST node to represent enum literals.
The `ExpressionVisitor` was updated to handle and map enum literals with Z3's enum sorts and constants, while the translator creates and caches these.
The `MethodsFirstChecker` and `RefinementTypeChecker` have also been modified to add into the global context instead of just local and to recognize field accesses and annotate them with the proper refinement metadata for type checking

A set of tests was added to verify the valid/invalid behaviour of enum refinements, both `Refinement` and `StateRefinement`.


liquidjava-verifier/src/main/java/liquidjava/smt/TranslatorContextToZ3.java
```diff
@@ -48,7 +48,12 @@ private static Expr<?> getExpr(Context z3, String name, CtTypeReference<?> type)
         case "long", "java.lang.Long" -> z3.mkRealConst(name);
         case "float", "double", "java.lang.Float", "java.lang.Double" -> (FPExpr) z3.mkConst(name, z3.mkFPSort64());
         case "int[]" -> z3.mkArrayConst(name, z3.mkIntSort(), z3.mkIntSort());
-        default -> z3.mkConst(name, z3.mkUninterpretedSort(typeName));
+        case "java.lang.Enum" -> z3.mkIntConst(name);
+        default -> {
+            if (type.isEnum())
+                yield z3.mkIntConst(name);
+            yield z3.mkConst(name, z3.mkUninterpretedSort(typeName));
+        }
```
rcosta358: Converting the enums to integers is makes sense, but we can improve it. Instead, we can use the Z3 `EnumSort`, so the error message and the counterexample shows the enum names instead of the integers they represent.

rcosta358 (changes requested):
Nice work! Left some feedback about the parsing, representation, scoping and translation of enums.
Additionally, we need more tests to verify the correctness of this implementation.

rcosta358 (commented):
Left some minor comments.

CatarinaGamboa (changes requested):
Sorry for the delay @cheestree 
I left some comments to improve readability, but it seems to work well, so great! Let me know if you have questions on the comments

rcosta358 (approved):
Great! Left some minor suggestions.

cheestree (commented):
Changed the method name but not when called.

CatarinaGamboa (approved):
Great! I think we can merge it!

Issue #134, opened 2025-12-12
Overloading not accounting for arguments type for state transitions


CatarinaGamboa:
I think there's a bug in selecting the correct method to check when there are two methods with the same number of arguments but different types.

This was the example where I encountered the issue.

This class has two constructors with only 1 argument, but they go to different states. With the first, it's okay to call `initCause`, with the second, it is not.

```java
@StateSet({"start", "hasMessage", "hasCause"})
@ExternalRefinementsFor("java.lang.Throwable")
public interface ThrowableRefinements {

    // ##### Constructors #######
    @StateRefinement(to="hasMessage(this)")
    public void Throwable(String message);

    @StateRefinement(to="hasCause(this)")
    public void Throwable(Throwable cause);

    @StateRefinement(from="!hasCause(this)", to="hasCause(this)")
    public Throwable initCause(Throwable cause);
}
```

Tests where we can see the issue:
```java
public class Test{
    void test3(){
        Throwable t = new Throwable("Example");
        t.initCause(null);
        t.getCause();
    }

    void test4(){
        Throwable originT = new Throwable();
        Throwable t = new Throwable(originT); // should be an error but its not
        t.initCause(null);
        t.getCause();
    }

}
```
Now, what leads me to think this is an issue in selecting the methods is that we get errors in both test cases if:
- we remove the constructor with string
- change the order of the two constructors

PR #172, opened 2026-03-10
Logic for method overload with different types

CatarinaGamboa:
This PR fixes issue #134 .
We now account for the types of arguments when identifying the function that is called.
The main change is `checkInvocationRefinements` now has another parameter for the param types, and we change all its calls to capture the arguments' types and call the correct function.


CatarinaGamboa:
@copilot review it

liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/general_checkers/MethodsFunctionsChecker.java
```diff
@@ -247,38 +264,62 @@ private void searchMethodInLibrary(CtExecutableReference<?> ctr, CtInvocation<?>
...
+            checkInvocationRefinements(invocation, invocation.getArguments(), invocation.getTarget(), name, ctype,
+                    paramTypes);
             return;
         }
         if (qualifiedSignature != null) {
             String completeName = String.format("%s.%s", ctype, name);
-            if (rtc.getContext().getFunction(completeName, ctype, argSize) != null) {
+            f = rtc.getContext().getFunction(completeName, ctype, paramTypes);
+            if (f == null)
+                f = rtc.getContext().getFunction(completeName, ctype, argSize);
```
rcosta358:
```suggestion
```

PR #187, opened 2026-03-19
CLI flags

cheestree:
Closes #175 . This PR adds support for the `--debug` and `--version` flags as well as future implementation of new ones via the use of the `picocli` library.


rcosta358 (approved):
LGTM. Can you show us example usages of each option?

CatarinaGamboa (changes requested):
grest! but lets move the main print to another place to have more info

alcides (approved):
LGTM

PR #190, opened 2026-03-24
Add Missing `this` Binding + Method Checking Ordering

### LINKED ISSUE #78: Unexpected Error When Combining Refinement and StateRefinement
In the following example, both the state refinement and the parameter refinement are individually valid, but when used together, the type checker returns the error `Variable 'this' not found`, which is probably a side effect of this bug.

```java
@StateRefinement(from="downloading(this)", to="progress(this) == percentage")
public void updateProgress(@Refinement("percentage > progress(this)") int percentage) {} // Variable 'this' not found
```

By putting all the logic in the `@StateRefinement`, it works:

```java
@StateRefinement(from="downloading(this) && percentage > progress(this)", to="progress(this) == percentage")
public void updateProgress(int percentage) {} // no error
```

By removing either one of them, we also don't get an error:

```java
public void updateProgress(@Refinement("percentage > progress(this)") int percentage) {} // no error

// or

@StateRefinement(from="downloading(this)", to="progress(this) == percentage")
public void updateProgress(int percentage) {} // no error
```

There seems to be some kind of conflict between the two.

rcosta358:
Closes #78.

### Add Missing `this` Binding

Fixes the error `Variable 'this' not found` when a method combined a state refinement with a parameter refinement that referenced `this`. Fixed this by adding `this` to the context in `AuxStateHandler.createStatePredicate`.

### Method Checking Ordering

This PR also fixes invocation parameter refinement checking to use the receiver’s pre-call state instead of its post-call state, which means the order of checking a parameter is now check **pre-conditions -> check parameter refinements -> update with post-conditions**. This keeps the existing state transition behavior, but delays applying it until after parameter validation.

#### Why?

In the following example, in the `update` method, the parameter refinement `percentage > progress(this)` should be checked against the receiver's state before the call. Instead, the verifier was applying the receiver state transition first and only then validating the parameter refinement, which meant the condition was always false (`x > x => false` ).

```java
@Ghost("int progress")
@StateSet({"created", "downloading", "completed"})
public class Downloader {

    @StateRefinement(to="created(this) && progress(this) == 0")
    public Downloader() {}

    @StateRefinement(from="created(this) && progress(this) == 0", to="downloading(this) && progress(this) == 0")
    public void start() {}

    @StateRefinement(from="downloading(this)", to="downloading(this) && progress(this) == percentage")
    public void update(@Refinement("percentage > progress(this)") int percentage) {}

    @StateRefinement(from="downloading(this) && progress(this) == 100", to="completed(this)")
    public void finish() {}
}
```


CatarinaGamboa (changes requested):
Let's just add some tests that actually exercise the state, along with the ghost just to check, that verification actually supports this

CatarinaGamboa (approved):
LGTM. Minor comment: in this example, we end up losing the information about `progress` at the end since it is not in the `to` state, but I think it still exemplifies the idea

rcosta358:
You're right, but since the download is complete the `progress` is irrelevant.

PR #199, opened 2026-03-30
Refactor "Multiple" API Annotations

rcosta358:
This PR cleans up the API by moving the repeatable container annotations for `Ghost`, `RefinementAlias`, `RefinementPredicate`, `StateRefinement`, and `StateSet` inside their corresponding annotation types. This way, the internal `Multiple` annotations are not suggested in the autocomplete and used by mistake.


CatarinaGamboa (approved):
Oh this works? Amazing 🥳

PR #200, opened 2026-03-31
Expression Pretty Printing

rcosta358:
This PR adds a formatter for expressions. Parentheses are only added when needed to preserve operator precedence and associativity, so expressions stay unambiguous without adding extra noise. Also simplifies the variable formatter and adds unit tests.


CatarinaGamboa (approved):
niceee

PR #201, opened 2026-04-01
Special Variable Substitution

rcosta358:
This PR changes the substitution of special variables so `#ret_n` is never replaced by another variable and `#fresh_n` is never used as the replacement for some other variable. 

#### Before

```
Refinement Error: x == x is not a subtype of #ret⁰ > 0
```

#### After

```
Refinement Error: #ret⁰ == x is not a subtype of #ret⁰ > 0
```


liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
```diff
@@ -54,16 +54,17 @@ private static void resolveRecursive(Expression exp, Map<String, Expression> map
                 map.put(var.getName(), left.clone());
             } else if (left instanceof Var leftVar && right instanceof Var rightVar) {
                 // to substitute internal variable with user-facing variable
-                if (leftVar.isInternal() && !rightVar.isInternal()) {
+                if (isInternal(leftVar) && !isInternal(rightVar) && !isReturnVar(leftVar)) {
```
CatarinaGamboa, line 57: could we have like a `isKeywordVar` or `isSpecialVar` to cover both `Fresh` and `Ret`?

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
```diff
@@ -144,4 +145,23 @@ private static boolean hasUsage(Expression exp, String name) {
...
+    private static boolean isInternal(Var var) {
+        return var.getName().startsWith("#");
+    }
+
+    private static boolean isReturnVar(Var var) {
+        return var.getName().startsWith("#ret_");
+    }
+
+    private static boolean isFreshVar(Var var) {
+        return var.getName().startsWith("#fresh_");
```
CatarinaGamboa, line 165: could these be dependent on the final strings where we define the format for fresh and ret instead of hard coded?

CatarinaGamboa (approved):
lgtm, check the minor comments i left

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
rcosta358: Technically yeah, but not cleanly, since the formats are `#ret_%d` and `#fresh_%d` and we only want the prefixes. It would only work with regexes, not string formats.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
rcosta358: We could, but since we need the `isReturnVar` independently, I think it makes more sense have those separate.

PR #202, opened 2026-04-02
Unsatisfiable Refinement Warning

rcosta358:
This PR adds a warning for unsatisfiable, self-contradictory refinements. It also avoids re-running the same satisfiability check for the same element across verifier passes.

### Examples

image


liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/TypeChecker.java
```diff
@@ -87,11 +94,48 @@ public Optional<Predicate> getRefinementFromAnnotation(CtElement element) throws
...
                         ref.get());
             }
+            if (!Boolean.TRUE.equals(element.getMetadata(Keys.REFINEMENT_SAT_CHECK)))
+                checkRefinementSatisfiability(ref.get(), p, element);
             constr = Optional.of(p);
         }
         return constr;
     }
 
+    private void checkRefinementSatisfiability(String refinement, Predicate predicate, CtElement element) {
```
CatarinaGamboa, line 110: javadoc plss

liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/TypeChecker.java
```diff
@@ -87,11 +94,48 @@ public Optional<Predicate> getRefinementFromAnnotation(CtElement element) throws
...
+        context.enterContext();
+        try {
+            Predicate p = new Predicate();
+            CtTypeReference<?> annotationType = element instanceof CtTypedElement<?> typedElement
+                    ? typedElement.getType() : null;
+            if (annotationType != null && !context.hasVariable(Keys.WILDCARD))
+                context.addVarToContext(Keys.WILDCARD, annotationType, p, element);
```
CatarinaGamboa, line 117: I'm a bit confused, why are we adding the vars to context here? because this happens before we add them to context?

CatarinaGamboa (commented):
I think I need some description of our approach on this. So we check satisfiability every time that we try to get the value of a `@Refinement` (and skip if it already has it). But why do we need to add the vars to context? what if a refinements mentions several params that are not yet in context?

rcosta358:
We do the satisfiability check once per element inside a temporary scope. We add names like `_`, the variable itself, `return`, and `this` to that scope because they are not in the context yet. If the refinement mentions other params that are still not in context, the satisfiability check just fails and we ignore it.

liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/TypeChecker.java
rcosta358: Yep.

liquidjava-verifier/src/main/java/liquidjava/smt/SMTEvaluator.java
```diff
@@ -48,4 +48,18 @@ public SMTResult verifySubtype(Predicate subRef, Predicate supRef, Context conte
...
         return SMTResult.ok();
     }
+
+    public boolean isUnsatisfiable(Predicate predicate, Context context) throws Exception {
+        try {
+            Expression exp = predicate.getExpression();
+            try (TranslatorToZ3 tz3 = new TranslatorToZ3(context)) {
+                ExpressionToZ3Visitor visitor = new ExpressionToZ3Visitor(tz3);
+                Expr<?> e = exp.accept(visitor);
+                Solver solver = tz3.makeSolverForExpression(e);
```
CatarinaGamboa, line 58: just to check if we dont find a given var this throws some exception? or just silently send a false in the next line?

CatarinaGamboa (approved):
lgtm, left a question

liquidjava-verifier/src/main/java/liquidjava/smt/SMTEvaluator.java
rcosta358: It silently fails and nothing happens. The worst case scenario is not displaying the warning when we should.

PR #203, opened 2026-04-06
Fix Wildcard Imports in External Projects

rcosta358:
This PR is a follow-up on #192, which only fixed the issue for examples in the current project. In external projects that used `import liquidjava.specification.*`, Spoon would mis-resolve LiquidJava annotations to the current package, for example `com.example.Refinement` instead of `liquidjava.specification.Refinement`.

To fix this, added the `liquidjava-api` to Spoon's source classpath in `CommandLineLauncher`, while keeping `noClasspath` mode enabled. The path is obtained from the loaded `Refinement` annotation class and points to the `liquidjava-api` artifact, which contains all LiquidJava annotations. This allows Spoon to resolve wildcard-imported LiquidJava annotations correctly in external projects.


CatarinaGamboa (approved):
who would know this sequence without ai 🤯

rcosta358:
I know right 😂

PR #204, opened 2026-04-06
Filter Duplicate Counterexample Assignments

rcosta358:
This PR improves counterexample filtering by only keeping assignments for variables that appear in the found expression and whose exact assignments are not already present there. This avoids relying on a full string comparison between the counterexample and the found expression, which could fail when the same assignments appeared in a different order.

#### Before

image

#### After

image


liquidjava-verifier/src/main/java/liquidjava/diagnostics/errors/RefinementError.java
```diff
@@ -46,16 +47,17 @@ public String getCounterExampleString() {
 
         List<String> foundVarNames = new ArrayList<>();
         found.getValue().getVariableNames(foundVarNames);
+        List<String> foundAssignments = found.getValue().getConjuncts().stream().map(Expression::toString).toList();
         String counterexampleString = counterexample.assignments().stream()
-                // only include variables that appear in the found value
-                .filter(a -> foundVarNames.contains(a.first()))
+                // only include variables that appear in the found value and are not already fixed there
+                .filter(a -> foundVarNames.contains(a.first())
+                        && !foundAssignments.contains(a.first() + " == " + a.second()))
```
CatarinaGamboa: What if `a.second()`  is another variable? Or are we already sure that we are in a simplified version where thats not possible

rcosta358: I'm not sure, but I'm guessing that Z3 always returns the most simplified form, meaning the RHS would be a value and not another variable. If that's not the case for some weird example, the worst thing that can happen is showing an extra assignment in the counterexample (which is better than incorrectly hiding it).

CatarinaGamboa (approved):
lgtm

PR #205, opened 2026-04-07
Refactor Refinement Predicate

rcosta358:
This PR refactors the usage and semantics of the `@RefinementPredicate` annotation.
It has two main changes:
- Removes the `ghost` and `type` keywords from the grammar, as they were redundant and provided no practical benefit
- Changes the annotation target from methods/constructors to classes/interfaces, since it adds the ghost globally and not on a method-level scope

These changes simplify the annotation model and make its usage more intuitive and consistent.


liquidjava-api/src/main/java/liquidjava/specification/RefinementPredicateMultiple.java
```diff
@@ -11,7 +11,7 @@
  * @author Catarina Gamboa
  */
 @Retention(RetentionPolicy.RUNTIME)
-@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@Target({ElementType.TYPE})
 public @interface RefinementPredicateMultiple {
```
CatarinaGamboa: you can use the same trick as you used for the other ones

CatarinaGamboa (approved):
lgtm

liquidjava-api/src/main/java/liquidjava/specification/RefinementPredicateMultiple.java
rcosta358: Yeah, this branch didn't have the changes from #199.

PR #207, opened 2026-05-04
Support for class Constants 

CatarinaGamboa:
We did not support using static final fields from classes in predicates or using the values in Java code.
For example, both these cases would not have been possible before, and now are acceptable
```java
    static void clampInt(@Refinement("_ >= Integer.MIN_VALUE && _ <= Integer.MAX_VALUE") int x) {
    }

    public static void main(String[] args) {
        clampInt(0);
        clampInt(Integer.MAX_VALUE);
    }
```
  - Resolution follows Java scoping rules: fully-qualified name → compilation-unit imports (single + on-demand) → implicit java.lang, with reflection as the source-of-truth. User-defined enums/classes are left untouched for the existing SMT path.                      
  - Based on reflection, find the declaration and get the value assigned to it, and add it as a value.

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/Predicate.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/Predicate.java
@@ -98,6 +98,7 @@ public class Predicate {
             }
             if (StaticConstants.userTypeExists(en.getTypeName(), context))
                 continue; // likely a user-defined enum/class — let SMT translation handle it
+            // unresolvable reference — throw an error with a helpful message and import suggestion if possible
             SourcePosition pos = context == null ? null : Utils.getLJAnnotationPosition(context, en.toString());
             String suggested = StaticConstants.findImportCandidate(en.getTypeName(), en.getConstName(), context);
             String hint = suggested != null ? "add: import " + suggested + ";"
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/Predicate.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/Predicate.java
@@ -96,14 +96,16 @@ public class Predicate {
                 e = e.substitute(en, lit.getExpression());
                 continue;
             }
-            String suggested = StaticConstants.findJdkImportCandidate(en.getTypeName(), en.getConstName());
-            if (suggested != null) {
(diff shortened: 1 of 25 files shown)


liquidjava-verifier/src/main/java/liquidjava/rj_language/Predicate.java
```diff
@@ -75,9 +77,52 @@ public Predicate(String ref, CtElement element, String prefix) throws LJError {
         if (!(exp instanceof GroupExpression)) {
             exp = new GroupExpression(exp);
         }
+        exp = resolveStaticFinalConstants(exp, element);
```
CatarinaGamboa, line 81: > Do we need to call this all the time? 

it will just see if there are enums to resolve in a small expression, if there are no enums it doesnt do anything else, so its fine

CatarinaGamboa:
@rcosta358 I'm not sure how this affects error messages as it will appear directly with the value in the SMT, but there is no link between the var and its value so we might need a followup pr to handle that

rcosta358:
Can't we treat those as variables instead? That way we'll solve that for free.

CatarinaGamboa:
> Can't we treat those as variables instead? That way we'll solve that for free.

we can add a new expression with the equality instead of replacing it, is that what you mean?

rcosta358:
Yep, exactly, that way we can also expand them to their values in the simplification.

liquidjava-verifier/src/main/java/liquidjava/rj_language/ast/Enum.java
```diff
@@ -9,6 +9,13 @@ public class Enum extends Expression {
 
     private final String typeName;
     private final String constName;
+    /**
+     * If this {@code Type.CONST} reference resolved to a Java {@code static final} primitive/String constant, the
+     * corresponding RJ literal expression is stashed here so the SMT translator can emit a binding axiom
+     * ({@code Type.CONST == literalValue}) while preserving the symbolic name in the AST. {@code null} for user-defined
+     * enum constants and for unresolvable references.
+     */
+    private Expression resolvedLiteral;
```
CatarinaGamboa, line 18: We keep the resolved value here instead of substituting it

rcosta358 (approved):
LGTM. Also fixed some issues, namely improving the error messages related to `static final` fields, referencing local `static final` fields (which don't need to be imported), and resolving `static final char` fields.

PR #209, opened 2026-05-05
Fix Operator Precedence

rcosta358:
This PR fixes operator precedence in the LiquidJava refinements, where e.g. the operators `+` and `-` had the same precedence as `*`, `/`, and `%`. This could lead to unexpected bugs and forced users to explicitly group expressions in parentheses.

Also, the grammar rejected some valid nested boolean expressions, such as `"_ == (true || false && false)"`, because `&&` and `||` were not allowed in the `exp` rule.

Finally, the parser could accept only a valid prefix of the expression. For example, the refinement `true false` would be accepted as `true` instead of throwing a syntax error, silently ignoring `false`. 

### Changes

The grammar now defines expression parsing in a single left-recursive `pred` rule, making the precedence order explicit, with the last two being right associative:

`!` → `* / %` → `+ -` → `== != >= > <= <` → `&&` → `||` → `-->` → `? :`

To make sure the whole refinement string is consumed, `prog: start | ;` was changed to to `prog: start EOF;` and `parser.start()` to `parser.prog()`.

### Examples

```java
@Refinement("_ == 1 + 2 * 0")
int x = 1;
// Before: Refinement Error: #x⁰ == 1 is not a subtype of #x⁰ == 0
// After: No error
```

```java
@Refinement("_ == (true || false && false)")
// Before: Syntax Error
// After: No error
```

```java
@Refinement("true false")
// Before: No error
// After: Syntax Error
```

### DIFF
--- a/liquidjava-verifier/src/main/antlr4/rj/grammar/RJ.g4
+++ b/liquidjava-verifier/src/main/antlr4/rj/grammar/RJ.g4
@@ -1,41 +1,31 @@
 grammar RJ;
 
 
-prog: start | ;
+prog: start EOF;
 start:
 		pred	#startPred
 	|	alias	#startAlias
 	|	ghost	#startGhost
 	;
 
+//----------------------- Predicates -----------------------
 
 pred:
-		'(' pred ')'				#predGroup
-	|	'!' pred					#predNegate
-	|	pred LOGOP pred 			#predLogic
-	|	pred '?' pred ':' pred		#ite
-	|	exp							#predExp
+		'-' pred					#opMinus
+	|	'!' pred					#opNot
+	|	'(' pred ')'				#predGroup
+	|	literalExpression			#opLiteral
+	|	pred MULOP pred				#opArith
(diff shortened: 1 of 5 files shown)


liquidjava-verifier/src/main/antlr4/rj/grammar/RJ.g4
```diff
@@ -1,41 +1,31 @@
...
+		'-' pred					#opMinus
+	|	'!' pred					#opNot
+	|	'(' pred ')'				#predGroup
+	|	literalExpression			#opLiteral
+	|	pred MULOP pred				#opArith
+	|	pred ('+'|'-') pred			#opArith
+	|	pred BOOLOP pred			#expBool
+	|	pred '&&' pred 				#predLogic
+	|	pred '||' pred 				#predLogic
+	|	<assoc=right> pred '-->' pred #predLogic
```
CatarinaGamboa, line 23: ok didnt know you could use <assoc=right>

CatarinaGamboa (approved):
ok this looks good

PR #214, opened 2026-05-09
Operator Assignment Support

rcosta358:
## Description
This PR adds support for Java operator assignments such as `+=`, `-=`, `*=`, `/=`, and `%=` in the refinement checking.
It works by modeling operator assignments as equivalent simple assignments: `x += y` → `x = x + y`.

## Example
```java
@Refinement("_ > 0")
int plus(@Refinement("_ >= 0") int x) {
    x += 1; // x is now > 0
    return x;
}
```
Before we had an error in the return statement because the `x += 1` was ignored. Now we don't.

## Related Issue
None.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added/updated tests under `liquidjava-example/src/main/java/testSuite/` (`Correct*` / `Error*`)

### DIFF
--- a/liquidjava-example/src/main/java/testSuite/CorrectOperatorAssignments.java
+++ b/liquidjava-example/src/main/java/testSuite/CorrectOperatorAssignments.java
@@ -1,0 +1,47 @@
+package testSuite;
+
+import liquidjava.specification.Refinement;
+
+public class CorrectOperatorAssignments {
+
+    @Refinement("_ > 0")
+    int plus(@Refinement("_ >= 0") int x) {
+        x += 1; // x is now > 0
+        return x;
+    }
+
+    @Refinement("_ > 0")
+    int plusInvocation(@Refinement("_ >= 0") int x) {
+        x += one(); // x is now > 0
+        return x;
+    }
+
+    @Refinement("_ == 1")
+    int one() {
+        return 1;
+    }
+
+    @Refinement("_ < 0")
+    int minus(@Refinement("_ <= 0") int x) {
+        x -= 1; // x is now < 0
+        return x;
+    }
+
+    @Refinement("_ >= 0")
+    int multiply(int x) {
+        x *= x; // x is now >= 0
+        return x;
+    }
+
+    @Refinement("_ <= 1")
+    int divide(@Refinement("0 <= _ && _ <= 3") int x) {
+        x /= 2; // x is now <= 1
+        return x;
+    }
+
+    @Refinement("_ == 0 || _ == 1")
+    int remainder(@Refinement("_ >= 0") int x) {
+        x %= 2; // x is now == 0 || x is now == 1
+        return x;
+    }
+}
--- a/liquidjava-example/src/main/java/testSuite/CorrectOperatorAssignments.java
+++ b/liquidjava-example/src/main/java/testSuite/CorrectOperatorAssignments.java
@@ -44,4 +44,45 @@ public class CorrectOperatorAssignments {
         x %= 2; // x is now == 0 || x is now == 1
         return x;
     }
+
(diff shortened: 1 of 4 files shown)


liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/general_checkers/OperationsChecker.java
```diff
@@ -280,6 +296,48 @@ private Predicate getOperationRefinementFromExternalLib(CtInvocation<?> inv) thr
...
+            return getCurrentVariableValue(name);
+        } else if (element instanceof CtBinaryOperator<?> binaryOperator) {
+            Predicate left = getOperatorAssignmentRefinement(binaryOperator.getLeftHandOperand());
+            Predicate right = getOperatorAssignmentRefinement(binaryOperator.getRightHandOperand());
+            return Predicate.createOperation(left, getOperatorFromKind(binaryOperator.getKind()), right);
+        } else if (element instanceof CtLiteral<?> literal) {
+            if (literal.getValue() == null)
+                throw new CustomError("Null literals are not supported", literal.getPosition());
+            return new Predicate(literal.getValue().toString(), element);
+        }
```
CatarinaGamboa, line 334: What if it is another kind of expression? we need more fallbacks.
Counter-example (with your example class):
```
    int t3(@Refinement("_ >= 0") int x) {
        @Refinement("_ == 10 || _ == 11")
        int y = 10;
        y += remainder(x); // x is now >= 6
        return x;
    }
```
outputs
```
Error: class com.microsoft.z3.BoolExpr cannot be cast to class com.microsoft.z3.ArithExpr (com.microsoft.z3.BoolExpr and com.microsoft.z3.ArithExpr are in unnamed module of loader org.codehaus.mojo.exec.URLClassLoaderBuilder$ExecJavaClassLoader @cb7fa71)
```

liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/RefinementTypeChecker.java
```diff
@@ -186,10 +187,23 @@ public <T> void visitCtThisAccess(CtThisAccess<T> thisAccess) {
...
 
-    @SuppressWarnings("unchecked")
     @Override
     public <T, A extends T> void visitCtAssignment(CtAssignment<T, A> assignment) throws LJError {
         super.visitCtAssignment(assignment);
+        visitAssignment(assignment);
+    }
+
+    @Override
+    public <T, A extends T> void visitCtOperatorAssignment(CtOperatorAssignment<T, A> assignment) {
```
CatarinaGamboa: throws LJError?

liquidjava-example/src/main/java/testSuite/CorrectOperatorAssignments.java
```diff
@@ -0,0 +1,47 @@
...
+        x /= 2; // x is now <= 1
+        return x;
+    }
+
+    @Refinement("_ == 0 || _ == 1")
+    int remainder(@Refinement("_ >= 0") int x) {
+        x %= 2; // x is now == 0 || x is now == 1
+        return x;
+    }
+}
```
CatarinaGamboa, line 88: Can we also get some negative cases for examples?

CatarinaGamboa (changes requested):
Needs to cover more operations on the RHS, and more examples that fail the refinements.
Also, there are more similar ops that we are not targetting right?  like bitwise `&=`, ..., we don't need but it should be a todo or maybe we should throw an error or something say we dont support them

rcosta358:
You're right. Will fix it.

CatarinaGamboa (approved):
lgtm

PR #215, opened 2026-05-09
Propagate Expressions

rcosta358:
## Description
This PR enables the propagation of expressions, instead of only literals or other variables.
With this, we can propagate all the following: `var == literal`, `var == var`, and now `var == expression`.
For example, the assignment `x == y + 1` now adds the mapping `x → y + 1`, allowing the RHS to be propagated wherever `x` appears.

#### Before

```
Refinement Error: #ret³ == y² - 2 && y² == x + 1 is not a subtype of #ret³ > 0
```

#### After

```
Refinement Error: #ret³ == x - 1 is not a subtype of #ret³ > 0
```

## Related Issue
None.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added tests in `ExpressionSimplifierTest`

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/ExpressionFolding.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/ExpressionFolding.java
@@ -146,6 +146,10 @@ public class ExpressionFolding {
                 return new ValDerivationNode(res, new BinaryDerivationNode(leftNode, rightNode, op));
         }
 
+        ValDerivationNode adjacentConstants = foldAdjacentIntegerConstants(leftNode, rightNode, op);
+        if (adjacentConstants != null)
+            return adjacentConstants;
+
         // no folding
         DerivationNode origin = (leftNode.getOrigin() != null || rightNode.getOrigin() != null)
                 ? new BinaryDerivationNode(leftNode, rightNode, op) : null;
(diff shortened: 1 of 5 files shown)


liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/ExpressionFolding.java
```diff
@@ -243,4 +247,32 @@ private static ValDerivationNode foldIte(ValDerivationNode node) {
...
+            return null;
+        if (!(leftNode.getValue()instanceof BinaryExpression leftBinary))
+            return null;
+        if (!"+".equals(leftBinary.getOperator()) && !"-".equals(leftBinary.getOperator()))
+            return null;
+        if (!(leftBinary.getSecondOperand()instanceof LiteralInt leftLiteral))
+            return null;
+
+        int signedLeft = "+".equals(leftBinary.getOperator()) ? leftLiteral.getValue() : -leftLiteral.getValue();
+        int signedRight = "+".equals(op) ? rightLiteral.getValue() : -rightLiteral.getValue();
```
CatarinaGamboa, line 265: overflows? maybe just an if checking its fine

CatarinaGamboa (approved):
Minor changes.
For you thesis, you should plan to write a pseudo-code algorithm of how simplification of expressions works with all the steps and additions you have been making.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/ExpressionFolding.java
```diff
@@ -243,4 +247,32 @@ private static ValDerivationNode foldIte(ValDerivationNode node) {
...
-}
\ No newline at end of file
+
+    private static ValDerivationNode foldAdjacentIntegerConstants(ValDerivationNode leftNode,
+            ValDerivationNode rightNode, String op) {
+        if (!"+".equals(op) && !"-".equals(op))
+            return null;
+        if (!(rightNode.getValue()instanceof LiteralInt rightLiteral))
+            return null;
+        if (!(leftNode.getValue()instanceof BinaryExpression leftBinary))
```
CatarinaGamboa, line 257: nit spacee between method and `instanceof`

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/ExpressionFolding.java
rcosta358: Good catch. I think that can happen in more places, not just here. Will fix it in a follow-up PR.

PR #216, opened 2026-05-09
Propagate Function Invocations

rcosta358:
## Description
This PR allows function invocations to be on the LHS of mappings (instead of just variables).
This works by using the function string representation as a key in the variable resolver map, meaning they behave as if they were variables.
For example, `func(a) == func(b) && func(b) == 1` → `func(a) == 1`.

#### Before

```
State Refinement Error: Expected state size(stack²⁹⁵) > 0 but found size(stack²⁹⁵) == size(stack²⁹⁴) - 1 && size(stack²⁹⁴) == size(stack²⁹²) + 1 && size(stack²⁹²) == 0
```

#### After

```
State Refinement Error: Expected state size(stack²⁹⁴) > 0 but found size(stack²⁹⁴) == 0
```

## Related Issue
None.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added tests in `ExpressionSimplifierTest`

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
@@ -66,6 +66,8 @@ public class VariableResolver {
                     if (!isReturnVar(lowerVar) && !isFreshVar(higherVar))
                         map.putIfAbsent(lowerVar.getName(), higherVar.clone());
                 }
+            } else if (left instanceof Var var && !(right instanceof Var) && canSubstitute(var, right)) {
+                map.put(var.getName(), right.clone());
             }
         }
     }
@@ -124,7 +126,8 @@ public class VariableResolver {
         if (exp instanceof BinaryExpression binary && "==".equals(binary.getOperator())) {
             Expression left = binary.getFirstOperand();
             Expression right = binary.getSecondOperand();
-            if (left instanceof Var v && v.getName().equals(name) && right.isLiteral())
+            if (left instanceof Var v && v.getName().equals(name)
+                    && (right.isLiteral() || (!(right instanceof Var) && canSubstitute(v, right))))
                 return false;
             if (right instanceof Var v && v.getName().equals(name) && left.isLiteral())
                 return false;
@@ -163,5 +166,23 @@ public class VariableResolver {
 
     private static boolean isFreshVar(Var var) {
         return var.getName().startsWith("#fresh_");
+    }
+
+    private static boolean canSubstitute(Var var, Expression value) {
+        return !isReturnVar(var) && !isFreshVar(var) && !containsVariable(value, var.getName());
+    }
+
+    private static boolean containsVariable(Expression exp, String name) {
+        if (exp instanceof Var var)
+            return var.getName().equals(name);
+
+        if (!exp.hasChildren())
+            return false;
+
+        for (Expression child : exp.getChildren()) {
+            if (containsVariable(child, name))
+                return true;
+        }
+        return false;
     }
 }
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
@@ -26,7 +26,7 @@ public class VariableResolver {
         resolveRecursive(exp, map);
 
         // remove variables that were not used in the expression
-        map.entrySet().removeIf(entry -> !hasUsage(exp, entry.getKey()));
+        map.entrySet().removeIf(entry -> !hasUsage(exp, entry.getKey(), entry.getValue()));
 
         // transitively resolve variables
         return resolveTransitive(map);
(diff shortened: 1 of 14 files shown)


liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
```diff
@@ -45,33 +46,50 @@ private static void resolveRecursive(Expression exp, Map<String, Expression> map
...
-                map.put(var.getName(), right.clone());
+            return;
+        }
+        if (!"==".equals(op))
+            return;
+
+        Expression left = be.getFirstOperand();
+        Expression right = be.getSecondOperand();
+        String leftKey = substitutionKey(left);
+        String rightKey = substitutionKey(right);
```
CatarinaGamboa, line 57: Is using strings for substitution really the best approach?

liquidjava-verifier/src/test/java/liquidjava/rj_language/opt/ExpressionSimplifierTest.java
```diff
@@ -1133,4 +1134,28 @@ void testFoldsAdjacentIntegerConstantsInLeftAssociatedArithmetic() {
...
+        Expression sizeX2EqualsSizeX1Plus1 = new BinaryExpression(sizeX2, "==",
+                new BinaryExpression(sizeX1, "+", new LiteralInt(1)));
+        Expression sizeX1Equals0 = new BinaryExpression(sizeX1, "==", new LiteralInt(0));
+        Expression fullExpression = new BinaryExpression(sizeX3EqualsSizeX2Minus1, "&&",
+                new BinaryExpression(sizeX2EqualsSizeX1Plus1, "&&", sizeX1Equals0));
+
+        ValDerivationNode result = ExpressionSimplifier.simplify(fullExpression);
+
+        assertEquals("size(x3) == 0", result.getValue().toString());
+    }
```
CatarinaGamboa, line 612: we need a couple more tests for this, I'm not very convinced

CatarinaGamboa (changes requested):
Is using strings here the best idea? Couldn't we use the expressions?
Also, we need more examples.
I was wondering, for the test suite we have already, we dont check the simplification right? we only have tests for "if an error in that line of type T", should we plan for a startegy to also have more tests for simplification?

rcosta358:
The mapping we had was `variable name` → `expression`, so using the string representation of the function invocation as a "variable" made sense to me, and I think that's enough.
Using a mapping `expression` → `expression` seems a bit weird to me.
Regarding the tests, we have separate tests for the simplification, in `ExpressionSimplifierTest` and `VariableResolverTest`.

CatarinaGamboa:
> The mapping we had was `variable name` → `expression`, so using the string representation of the function invocation as a "variable" made sense to me, and I think that's enough. Using a mapping `expression` → `expression` seems a bit weird to me. Regarding the tests, we have separate tests for the simplification, in `ExpressionSimplifierTest` and `VariableResolverTest`.

Can we improve those tests to show the string?
like:
```java
  @Test
  void testFunctionInvocationEqualitiesPropagateTransitively() {
     String toTest = "size(x3) == size(x2) - 1 && size(x2) == size(x1) + 1 && size(x1) == 0";
      Expression expr = RefinementsParser.createAST(toTest, "");

      ValDerivationNode result = ExpressionSimplifier.simplify(expr);

      assertEquals("size(x3) == 0", result.getValue().toString());
  }
```
its pretty hard to understand the tests the way they are now

rcosta358:
Will do.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
```diff
@@ -45,33 +46,49 @@ private static void resolveRecursive(Expression exp, Map<String, Expression> map
...
+                boolean isLeftCounterLower = getCounter(leftVar) <= getCounter(rightVar);
+                Var lowerVar = isLeftCounterLower ? leftVar : rightVar;
+                Var higherVar = isLeftCounterLower ? rightVar : leftVar;
+                if (!isReturnVar(lowerVar) && !isFreshVar(higherVar))
+                    map.putIfAbsent(lowerVar.getName(), higherVar.clone());
             }
+        } else if (left instanceof Var var && canSubstitute(var, right)) {
+            map.put(var.getName(), right.clone());
+        } else if (left instanceof FunctionInvocation && !right.toString().contains(leftKey)) {
+            map.put(leftKey, right.clone());
```
CatarinaGamboa, line 80: this means we would not substitute something like:
`f(a) == ff(a) + b`

because the string on the right contains f(a) even though they are not related, right?
Ig better be conservative but we should be aware of this limitation

CatarinaGamboa (commented):
I would still prefer the map `Expr->Expr` than `String->Expr` I think it makes more sense, and we can use `equals` so I dont think it adds complexity and its more accurate.
But this should work for now if you want to take care of that in another PR, or ask for someone else's opinion

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
rcosta358: You're right, I'll replace that string check with an AST check.

CatarinaGamboa (approved):
This is a good fix for now, in the future we can think about refactoring to use expressions if we want.

alcides:
@rcosta358 you might want to read about equality saturation: https://cs.mcgill.ca/~cs520/2025/slides/eqsat.pdf

It is used inside SMT solvers, and we might want to apply it as a simplification procedure, tracking its changes.

rcosta358:
That might be a good idea. I'll keep that in mind in the future.

PR #217, opened 2026-05-10
Fix Expression Formatter

rcosta358:
## Description
This PR fixes the `ExpressionFormatter` so grouped expressions only have parentheses when needed for precedence or associativity. This simplifies expressions in the error messages and in the context debugger.

## Example
Before and after:
- `(1)` → `1`
- `(a > 0)` → `a > 0`
- `a && (b > 0)` → `a && b > 0`

Unchanged cases:
- `a - (b + c)`
- `x == (1 > 0)`
- `!(x > 0)`

## Related Issue
None.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Code refactoring

## Checklist
- [x] Added/updated in `ExpressionFormatterTest`

### DIFF
--- a/liquidjava-verifier/src/test/java/liquidjava/rj_language/ast/ExpressionFormatterTest.java
+++ b/liquidjava-verifier/src/test/java/liquidjava/rj_language/ast/ExpressionFormatterTest.java
@@ -1,6 +1,9 @@
 package liquidjava.rj_language.ast;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.List;
+
 import org.junit.jupiter.api.Test;
 
 class ExpressionFormatterTest {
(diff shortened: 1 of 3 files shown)


liquidjava-verifier/src/test/java/liquidjava/rj_language/ast/ExpressionFormatterTest.java
```diff
@@ -86,7 +105,7 @@ void formatsTernaryExpressions() {
         assertEquals("a ? b : c ? d : e", new Ite(new Var("a"), new Var("b"), nestedElse).toDisplayString());
         assertEquals("(a ? b : c) ? d : e",
                 new Ite(new GroupExpression(ite), new Var("d"), new Var("e")).toDisplayString());
-        assertEquals("a ? b : (c ? d : e)",
+        assertEquals("a ? b : c ? d : e",
```
CatarinaGamboa: in here I think we want the parenthesis to still be there, right? reading it, it could either be
`a ? b : c ? d : e`
a) `a ? b : (c ? d : e)`
b) `(a ? b : c) ? d : e`
Also we can improve the tree construction as well, but you can do that in a followup PR

CatarinaGamboa (changes requested):
I think we should keep the ite grouped expressions, maybe the --> as well. They would pass the parser correctly but its harder for developers to easily see the right associative rule - at least for me it is, what do you think?

rcosta358:
I think it's understandable without the parentheses but I get your point.
Will change it.

liquidjava-verifier/src/test/java/liquidjava/rj_language/ast/ExpressionFormatterTest.java
```diff
@@ -1,93 +1,88 @@
...
     }
 
     @Test
     void formatsInternalVariables() {
-        assertEquals("x", new Var("x").toDisplayString());
-        assertEquals("x²", new Var("#x_2").toDisplayString());
-        assertEquals("#fresh¹²", new Var("#fresh_12").toDisplayString());
-        assertEquals("#ret³", new Var("#ret_3").toDisplayString());
-        assertEquals("this#Class", new Var("this#Class").toDisplayString());
+        assertEquals("x", parse("x").toDisplayString());
```
CatarinaGamboa, line 23: didnt you want to try `assertEquals("x", parse("(x)").toDisplayString());`

CatarinaGamboa (approved):
I see what you did with the examples since the last ones added () on purpose. But now its strange that some tests are just 
`assertEquals("x", parse("x").toDisplayString());`
can we also add some with
`assertEquals("x", parse("(x)").toDisplayString());`

liquidjava-verifier/src/test/java/liquidjava/rj_language/ast/ExpressionFormatterTest.java
rcosta358: Yep that case is missing, thanks.

PR #219, opened 2026-05-12
Distinguish Definitions from Usages in Variable Resolver

rcosta358:
## Description
This PR fixes a problem in the simplification where value propagation could be skipped for expressions like `x == 1 && x == 2`. The resolver treated any equality involving `x` and a literal as the definition of `x`, even when only one of those equalities was the actual definition being propagated.

Now the resolver tracks the value that defined the substitution and uses it to distinguish the defining equality from later usages. For a direct contradiction like `x == 1 && x == 2`, either value can be picked for propagation: both expose the contradiction and fold to `false`.

### Before

```
mode⁷⁰ == 1 &&
(mode⁷⁰ == 2 ? compressionExplicit(param⁷¹) : startCompression(param⁷¹)) &&
state1(param⁷¹) == state1(param⁶⁹) &&
startTiling(param⁶⁹) &&
startCompression(param⁶⁹)
```

### After

```
startCompression(param⁷¹) &&
state1(param⁷¹) == state1(param⁶⁹) &&
startTiling(param⁶⁹) &&
startCompression(param⁶⁹)
```

## Related Issue
None.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added/updated tests in `ExpressionSimplifierTest` and `VariableResolverTest`

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
@@ -26,7 +26,7 @@ public class VariableResolver {
         resolveRecursive(exp, map);
 
         // remove variables that were not used in the expression
-        map.entrySet().removeIf(entry -> !hasUsage(exp, entry.getKey()));
+        map.entrySet().removeIf(entry -> !hasUsage(exp, entry.getKey(), entry.getValue()));
 
         // transitively resolve variables
         return resolveTransitive(map);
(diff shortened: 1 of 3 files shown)


CatarinaGamboa (approved):
Awesome! We still need something similar now to deal with final fields/enums right?

rcosta358:
Yeah we need to handle enums in the simplification process.
Will do it in a separate PR.

PR #220, opened 2026-05-13
Derive Developer Facing States

### LINKED ISSUE #206: State Names Displayed with Internal Identifiers
Sometimes there are states in the error messages that appear as `state_%d`.
These mean nothing to the user and makes it hard to understand the problem.

rcosta358:
## Description
This PR improves error messages by deriving developer-facing states from internal state equalities.
When an error contains an internal equality like `state1(new) == state1(old)` plus a known developer state for `old`, the diagnostic now displays the corresponding developer state for new and omits the internal equality once it has served that purpose.

This is for error messages only and does not change SMT verification behavior.

#### Before
```
startCompression(param⁷¹) &&
state1(param⁷¹) == state1(param⁶⁹) &&
startTiling(param⁶⁹) &&
startCompression(param⁶⁹)
```

#### After
```
startCompression(param⁷¹) &&
startTiling(param⁶⁹) &&
startCompression(param⁶⁹) &&
startTiling(param⁷¹)
```

## Related Issue
Closes #206.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
@@ -391,8 +391,10 @@ public class VCChecker {
         gatherVariables(found, lrv, mainVars);
         TranslationTable map = new TranslationTable();
         VCImplication foundState = joinPredicates(found, mainVars, lrv, map);
-        throw new StateRefinementError(position, expected.getExpression(),
-                foundState.toConjunctions().simplify(context).getValue(), map, customMessage);
+        Predicate foundConjunction = new Predicate(foundState.toConjunctions().simplify(context).getValue())
+                .addDerivedStateEqualities(context.getGhostStates());
+        throw new StateRefinementError(position, expected.getExpression(), foundConjunction.getExpression(), map,
+                customMessage);
     }
 
     protected void throwStateConflictError(SourcePosition position, Predicate expected) throws StateConflictError {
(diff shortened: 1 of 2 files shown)


alcides:
Exactly how do state1 and startTiling relate? It seems startTiling(x) is equals to state1(x) == SOME_CONSTANT_FOR_StartTiling. Is that the case?

CatarinaGamboa:
> Exactly how do state1 and startTiling relate? It seems startTiling(x) is equals to state1(x) == SOME_CONSTANT_FOR_StartTiling. Is that the case?

Internally, the states are encoded as an uninterpreted function with integer values
```
@StateSet("startTiling, explicitTiling, setTiling")
// internally, create state1( _ )
//                                == 0 : startTiling
//                                == 1 : explicitTiling
//                                == 2 : setTiling
```
Then this is very helpful when the state does not change in a function invocation we add to the context something like
```
state1(x1) == state1(x2)
```
but this means nothing to the user, so we want to show this using the developer facing states. So we need to apply the equality resolution before we do the state changes from `state1( _ ) == 0` : `startTiling`
Does that make sense?

alcides:
Yes, it does!

I think my concern is whether knowing that the state of x1 and x2 must be the same is important to understand the error message (instead of x1 being in state A and x2 also need to be in state A), since the error would persist even if it was state B.

CatarinaGamboa:
I think we can maybe just simplify when both sides are instances of the same variable.
Also, we should probably allow the click expand for these cases as well

CatarinaGamboa:
@rcosta358 when running the cli with the ImageWriteParam case I still get the previous error, can you check it?

rcosta358:
We need to merge main first

CatarinaGamboa (changes requested):
We need some tests to see if this is working

PR #223, opened 2026-05-14
Update Run Script

rcosta358:
## Description
This PR updates the `liquidjava` script to run `mvn compile`, but only when local sources have changed.
This makes the verification run against the latest changes while avoiding unnecessary compilations.

## Checklist
- [x] Updated docs/README if behavior or API changed


CatarinaGamboa (approved):
nicee

PR #226, opened 2026-05-14
Fix Usage Detection for Repeated Equalities

rcosta358:
## Description
This PR fixes the `VariableResolver.hasUsage` so a matching equality is only excluded as a definition when it appears in a definition under a conjunction. This means that repeated equalities inside nested expressions now count as usages, allowing substitutions to propagate and simplify correctly.

#### Before

Both `x == 1` count has definitions.

```java
x == 1 && (x == 1 ? a(y) : b(y)) // most simplified
```

#### After

The first one counts has a definition and the second one counts as an usage.

```java
a(y) // most simplified
```

## Related Issue
None.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added/updated tests in `ExpressionSimplifierTests` and `VariableResolverTests`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,26 +7,38 @@ Thanks for your interest in contributing! This guide covers the essentials.
 - Java 20+
 - Maven 3.6+
 
-## Build & test
+## Build and Run
+
+To build the project, run:
 
 ```bash
-mvn clean install   # build everything
-mvn test            # run the test suite
+mvn clean install
 ```
 
-Run a single test:
+Code formatting runs automatically via `formatter-maven-plugin` during the `validate` phase.
 
-```bash
-mvn -pl liquidjava-verifier -Dtest=TestExamples#testMultiplePaths test
-```
-
-Verify a single file from the CLI:
+To verify a single file from the CLI, run:
 
 ```bash
 ./liquidjava path/to/File.java
 ```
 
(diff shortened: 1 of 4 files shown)


CatarinaGamboa:
what do you mean by "a matching equality is only excluded as a definition when it appears in a definition under a conjunction"? 
It could be the opposite right `(x == 1 ? a(y) : b(y)) && x == 1 ` and it should still simplify
I'm not sure i get the idea

rcosta358:
Those two expressions are equivalent. It's not about the order of the conjunctions, it's about whether the equality appears as a top level conjunct or nested inside another expression.

In both examples, the outer `x == 1` treated as the definition, while the `x == 1` within the ite is treated as a usage.
In the previous logic both equalities would count as definitions, so `x` had no usages and the substitution mapping was removed.

CatarinaGamboa (changes requested):
Can we add a couple more tests to ensure similar but more complex expressions are working?
Maybe like:
`mode == 2 && other == 5 && ((mode == 2 && other == 5) ? explicit(param) : start(param))` -> `explicit(param)`
or
`"mode == 2 && other == 3 && (mode == 2 ? (other == 3 ? a(p) : b(p)) : c(p))"` -> `a(p)`

CatarinaGamboa (approved):
lgtm

PR #229, opened 2026-05-18
Store Derivation Nodes in State Refinement Errors

rcosta358:
This PR makes state refinement errors store `ValDerivationNode` values for expected and found expressions, instead of a plain `String` to allow expanding expressions in state refinement errors in the extension.


CatarinaGamboa (approved):
great!

PR #233, opened 2026-05-28
Refer return values in StateRefinements

CatarinaGamboa:
## Description
The previous approach to  add state dependent on return values was attached to the return refinements
```
    @Refinement("_ --> acceptCompression(this)")
    boolean canWriteCompressed();
```
However, this is not a great mental model since we should change object states only using `@StateRefinements`. Therefore, now we allow using StateRefinemets for this, and fix previous bugs with the previous logic.

```
@StateRefinement(from = "start(this)",
                    to   = "return ? acceptCompression(this) : start(this)")
boolean canWriteCompressed();
```

## Example
Now we can verify:
```
ImageWriteParam setCompressionPreferences() {
    ImageWriteParam param = new ImageWriteParam(Locale.getDefault());
    if (param.canWriteCompressed()) {
        param.setCompressionMode(ImageWriteParam.MODE_DEFAULT);
        param.setCompressionQuality(0.85f); // State Refinement Error
    }
    return param;
}
```
where before we would get an error because `param.canWriteCompressed()` implied a state that was in conflict with the previous state assigned in the constructor.

Added several tests for the feature.

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added/updated tests under `liquidjava-example/src/main/java/testSuite/` (`Correct*` / `Error*`)

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/general_checkers/MethodsFunctionsChecker.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/general_checkers/MethodsFunctionsChecker.java
@@ -307,7 +307,7 @@ public class MethodsFunctionsChecker {
                 AuxStateHandler.checkTargetChanges(rtc, f, target, map, invocation);
 
             // Expose `_ == returnViName` so the if-condition path variable ties to this return value.
-            Predicate emptyRef = returnViName != null
+            Predicate returnRef = returnViName != null
                     ? Predicate.createEquals(Predicate.createVar(Keys.WILDCARD), Predicate.createVar(returnViName))
                     : new Predicate();
             invocation.putMetadata(Keys.REFINEMENT, emptyRef);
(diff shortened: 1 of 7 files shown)


liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/general_checkers/MethodsFunctionsChecker.java
```diff
@@ -289,14 +289,28 @@ private Map<String, String> checkInvocationRefinements(CtElement invocation, Lis
...
         if (target != null)
             AuxStateHandler.prepareInvocationTarget(rtc, target, invocation);
 
         if (f.allRefinementsTrue()) {
             if (target != null)
                 AuxStateHandler.checkTargetChanges(rtc, f, target, map, invocation);
 
-            invocation.putMetadata(Keys.REFINEMENT, new Predicate());
+            // Expose `_ == returnViName` so the if-condition path variable ties to this return value.
+            Predicate emptyRef = returnViName != null
```
rcosta358:
```suggestion
            Predicate returnRef = returnViName != null
```

liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/general_checkers/MethodsFunctionsChecker.java
```diff
@@ -289,14 +289,28 @@ private Map<String, String> checkInvocationRefinements(CtElement invocation, Lis
...
+                    ? Predicate.createEquals(Predicate.createVar(Keys.WILDCARD), Predicate.createVar(returnViName))
+                    : new Predicate();
+            invocation.putMetadata(Keys.REFINEMENT, emptyRef);
```
rcosta358:
```suggestion
            invocation.putMetadata(Keys.REFINEMENT, returnRef);
```

rcosta358 (approved):
LGTM.

PR #234, opened 2026-05-29
Resolve bare #x / !#x conjuncts as boolean substitutions to trigger simplification

CatarinaGamboa:
## Description
Simplification was not triggered for conjunctions without literals like `!hasNext`, so we've added a step to resolve the variables to their boolean values.

## Example
The error messages would not simplify, so:
Before:
```
State Refinement Error: Expected state hasNextElem(it³¹) but found 
(hasNext³⁰ ? hasNextElem(it³¹) : empty(it³¹)) && index(it³¹) == 0 && size(it³¹) == 5 && maybeEmpty(it²⁸) && false == hasNext¹⁶ && !#fresh³² && #fresh³² == hasNext³⁰
45 |         while (true){
46 |             if(!it.hasNext()){
47 |                 sum += it.next(); // State Refinement Error
   |                        ^^^^^^^^^
48 |             } else {
49 |                 break;

```

Now:
```
State Refinement Error: Expected state hasNextElem(it³¹) but found 
empty(it³¹) && index(it³¹) == 0 && size(it³¹) == 5 && maybeEmpty(it²⁸) && false == hasNext¹⁶
45 |         while (true){
46 |             if(!it.hasNext()){
47 |                 sum += it.next(); // State Refinement Error
   |                        ^^^^^^^^^
48 |             } else {
49 |                 break;

```

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x ] Added/updated tests under `liquidjava-example/src/main/java/testSuite/` (`Correct*` / `Error*`)

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
@@ -9,6 +9,8 @@ import liquidjava.rj_language.ast.BinaryExpression;
 import liquidjava.rj_language.ast.Enum;
 import liquidjava.rj_language.ast.Expression;
 import liquidjava.rj_language.ast.FunctionInvocation;
+import liquidjava.rj_language.ast.LiteralBoolean;
+import liquidjava.rj_language.ast.UnaryExpression;
 import liquidjava.rj_language.ast.Var;
 
 public class VariableResolver {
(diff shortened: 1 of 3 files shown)


liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VariableResolver.java
```diff
@@ -40,6 +42,18 @@ public static Map<String, Expression> resolve(Expression exp) {
      * @param map
      */
     private static void resolveRecursive(Expression exp, Map<String, Expression> map) {
+        // Internal-var conjuncts assert truth: `#x` ⇒ `#x → true`, `!#x` ⇒ `#x → false`. Restricted to
```
rcosta358, line 45: So if the expression is just `#x` or `!#x` they are simplified to `true`?  I don't think we want that.

rcosta358 (approved):
I think `#x` and `!#x` should only become boolean equalities when they are top level conjuncts and not when they are the entire expression or nested elsewhere.
This is because we are not correctly distinguishing definitions from usages.
We should fix this in a follow-up PR using the `VCImplications`, which would also fix the other problems with the simplification.
But this is okay for now.

CatarinaGamboa:
This is just another tape-in issue, lets improve the architecture beneath and then return to this issue

PR #236, opened 2026-05-30
Implicit Constructor Refinements

rcosta358:
## Description

This PR allows external refinement interfaces to omit constructor declarations while still correctly initializing typestate and ghost information, which previously was impossible:

> Constructors must always be present for typestate checking to work correctly, because they are the point where the initial state values are assigned. Otherwise, the initial values are not set and the verifier won’t be able to track the state of the object across method calls. [liquidjava-docs](https://liquid-java.github.io/liquidjava-docs/annotations/state-refinement/#state-initialization)

## Example

```java
@ExternalRefinementsFor("java.io.InputStreamReader")
@StateSet({"open", "closed"})
public interface InputStreamReaderRefinements {

    // public void InputStreamReader(InputStream in);

    @StateRefinement(from = "open(this)")
    int read();

    @StateRefinement(from = "open(this)", to = "closed(this)")
    void close();

    public static void main(String[] args) throws IOException {
        InputStreamReader isr = new InputStreamReader(System.in);
        isr.read();
        isr.close();
    }
}
```

Previously, `isr.read()` would fail because `isr` was initialized with an unknown state (`true`) when no constructor declaration was present.

Now, objects are automatically initialized with their default states, allowing the verification to work correctly even when constructors are omitted.

## Related Issue

None.

## Type of Change

* [x] Bug fix
* [x] New feature
* [ ] Documentation update
* [x] Code refactoring

## Checklist

* [x] Added/updated tests under `liquidjava-example/src/main/java/testSuite/` (`Correct*` / `Error*`)
* [x] `mvn test` passes locally
* [ ] Updated docs/README if behavior or API changed

### DIFF
--- a/liquidjava-example/src/main/java/testSuite/classes/input_stream_reader_no_constructor_correct/InputStreamReaderRefinements.java
+++ b/liquidjava-example/src/main/java/testSuite/classes/input_stream_reader_no_constructor_correct/InputStreamReaderRefinements.java
@@ -1,0 +1,15 @@
+package testSuite.classes.input_stream_reader_no_constructor_correct;
+
+import liquidjava.specification.*;
+
+// https://docs.oracle.com/javase/7/docs/api/java/io/InputStreamReader.html
+@ExternalRefinementsFor("java.io.InputStreamReader")
+@StateSet({"open", "closed"})
+public interface InputStreamReaderRefinements {
+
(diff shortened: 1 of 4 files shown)


CatarinaGamboa (approved):
lgtm. We need to change the docs to reflect this as you mentioned in the PR description

PR #243, opened 2026-06-02
Add -a/--all flag to show all SMT verification conditions

### LINKED ISSUE #240: Add a flag -a to show all verification conditions sent to the smt solver
**Problem**
When there is an error we can use `-d` to get more info. But when there isn't an error, and there should be, we also should be able to see more information than just "Correct! Passed verification" if we want to.

**Proposed solution**
Similar to debug logger, add info for every smt query that has been done.

**Alternatives considered**
Always showing everything is too much, using interactive debug is the current option but not the best

CatarinaGamboa:
## Description
Closes #240 

Now we show all the VCs that were sent to the smt solver even if there are no errors.

## Example
Screenshot 2026-06-02 at 3 56 03 PM

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/api/CommandLineArgs.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/api/CommandLineArgs.java
@@ -22,6 +22,9 @@ public class CommandLineArgs {
     @Option(names = { "-d", "--debug" }, description = "Enable debug mode for more detailed output")
     public boolean debugMode;
 
+    @Option(names = { "-a", "--all-context" }, description = "Print the full context after every checked element")
+    public boolean printContext;
+
     @Option(names = { "-lsp", "--language-server" }, description = "Enable language server mode for editor support")
     public boolean lspMode;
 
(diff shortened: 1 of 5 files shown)


liquidjava-verifier/src/main/java/liquidjava/api/CommandLineArgs.java
```diff
@@ -22,6 +22,9 @@ public class CommandLineArgs {
     @Option(names = { "-d", "--debug" }, description = "Enable debug mode for more detailed output")
     public boolean debugMode;
 
+    @Option(names = { "-a", "--all" }, description = "Show every verification condition sent to the SMT solver")
+    public boolean showAllVCs;
```
rcosta358: Maybe rename to something like `debugAllMode`.

liquidjava-verifier/src/main/java/liquidjava/diagnostics/DebugLog.java
```diff
@@ -37,12 +37,62 @@ public static boolean enabled() {
         return CommandLineLauncher.cmdArgs.debugMode;
     }
 
+    /**
+     * Gate for the SMT-query trace (verification conditions fed to Z3 and their outcomes). Enabled by either
+     * {@code --debug} / {@code -d} or the lighter {@code --all} / {@code -a}, which shows every query without the rest
+     * of the debug output (e.g. simplification passes).
+     */
+    public static boolean smtEnabled() {
```
rcosta358: Maybe rename to `loggingEnabled`.

rcosta358 (approved):

CatarinaGamboa:
We need more information for the cases we don't even have verification starting when there was supposed to be some, I'll work on this next

PR #245, opened 2026-06-03
Fix VCImplication Ordering

rcosta358:
## Description
This PR fixes VC implication construction order so each forall is introduced before it appears in the scope of another implication.

## Example

```java
int y = 5;
@Refinement("_ > 0")
int x = y;
```

### Before

```
∀#x_1:int. #x_1 == #y_0 =>
∀#y_0:int. #y_0 == 5
```

### After

```
∀#y_0:int. #y_0 == 5 =>
∀#x_1:int. #x_1 == #y_0
```

## Related Issue
None.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
@@ -290,8 +290,8 @@ public class VCChecker {
         }
         VCImplication cSMT = new VCImplication(new Predicate());
         if (firstSi != null) {
-            lastSi.setNext(new VCImplication(expectedType));
             cSMT = firstSi.clone();
+            lastSi.setNext(new VCImplication(expectedType));
         }
         return cSMT;
     }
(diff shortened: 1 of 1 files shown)


liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
```diff
@@ -252,20 +252,6 @@ private VCImplication joinPredicates(Predicate expectedType, List<RefinedVariabl
 
         VCImplication firstSi = null;
         VCImplication lastSi = null;
-        // Check
-        for (RefinedVariable var : mainVars) { // join main refinements of mainVars
```
rcosta358, line 256: Change order so variables are introduced before other refinements use them.

liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
```diff
@@ -331,6 +330,8 @@ private void getVariablesFromContext(List<String> lvars, List<RefinedVariable> a
                 .filter(rv -> !allVars.contains(rv)).forEach(rv -> {
                     allVars.add(rv);
                     recAuxGetVars(rv, allVars);
+                    allVars.remove(rv);
+                    allVars.add(rv);
```
rcosta358, line 334: Mark `rv` as seen first to avoid recursive cycles, then move it after the variables it uses.

CatarinaGamboa (approved):
lgtm! This whole code is so ready for a revamp haha

PR #246, opened 2026-06-08
Add SimplifiedPredicate

rcosta358:
This PR adds the `SimplifiedPredicate` that extends the `Predicate` class and stores the original and simplified predicates and binders.


liquidjava-verifier/src/main/java/liquidjava/rj_language/ast/formatter/ExpressionFormatter.java
```diff
@@ -61,8 +61,10 @@ private String formatArguments(List<Expression> args) {
     }
 
     private Expression unwrapGroup(Expression expression) {
-        while (expression instanceof GroupExpression group)
-            expression = group.getExpression();
+        while (expression instanceof GroupExpression) {
+            if (expression instanceof GroupExpression group)
```
CatarinaGamboa: why this if?

liquidjava-verifier/src/main/java/liquidjava/rj_language/SimplifiedPredicate.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import spoon.reflect.reference.CtTypeReference;
+
+public class SimplifiedPredicate extends Predicate {
+
+    private final Predicate simplified;
```
CatarinaGamboa: couldnt `simplified` be the expression from the super? we can leave it separate if you want but it feels like we could just use the super.expression, no?

liquidjava-verifier/src/main/java/liquidjava/rj_language/SimplifiedPredicate.java
```diff
@@ -0,0 +1,109 @@
...
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        SimplifiedPredicate other = (SimplifiedPredicate) obj;
+        return getSimplifiedPredicate().equals(other.getSimplifiedPredicate()) && origin.equals(other.origin)
+                && binders.equals(other.binders);
+    }
+
+    public static class Binder {
```
CatarinaGamboa, line 70: Do we nee dthis to be public? if yes create another file with it, but maybe we dont? not sure

liquidjava-verifier/src/main/java/liquidjava/rj_language/SimplifiedPredicate.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import spoon.reflect.reference.CtTypeReference;
+
+public class SimplifiedPredicate extends Predicate {
+
+    private final Predicate simplified;
+    private final Predicate origin;
+    private final List<Binder> binders;
```
CatarinaGamboa, line 20: say what you mean by binders

liquidjava-verifier/src/main/java/liquidjava/rj_language/SimplifiedPredicate.java
```diff
@@ -0,0 +1,109 @@
+package liquidjava.rj_language;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import spoon.reflect.reference.CtTypeReference;
+
+public class SimplifiedPredicate extends Predicate {
```
CatarinaGamboa, line 17: actually add some javadoc exlaining this class nature and what it represents with an example

CatarinaGamboa (changes requested):
minor changes

liquidjava-verifier/src/main/java/liquidjava/rj_language/ast/formatter/ExpressionFormatter.java
rcosta358: No need.

liquidjava-verifier/src/main/java/liquidjava/rj_language/SimplifiedPredicate.java
rcosta358: Yep, good call.

liquidjava-verifier/src/main/java/liquidjava/rj_language/SimplifiedPredicate.java
rcosta358: Yeah we need it to be public, and it is currently accessed as `SimplifiedPredicate.Binder` which I think makes more sense since we don't want to use `Binder` in any other context, so I think it should stay nested. What do you think?

CatarinaGamboa: hum ok i see, we can work with that

liquidjava-verifier/src/main/java/liquidjava/processor/refinement_checker/VCChecker.java
CatarinaGamboa: rebase issue? 😕 ?

rcosta358: Fixed.

CatarinaGamboa (approved):
lgtm

PR #249, opened 2026-06-09
Add VC Substitution Simplification

rcosta358:
## Description
This PR adds VC simplification for binder equalities through `VCImplication` chains. When a binder refinement establishes a known value (constant or not), the simplifier rewrites later refinements with that value and removes the substitution source from the VC chain preserving its origin and substituted binders.

It also adds focused unit tests and property-based tests using JUnit Quickcheck to randomly generate VCs and assert if the simplified VC is equivalent to the original.

## Example
```java
∀x:int. x == 3 --> x > 0

is simplified to

3 > 0

with origin ∀x:int. x > 0
```

## Related Issue
None.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added/updated tests


liquidjava-verifier/src/main/java/liquidjava/processor/VCImplication.java
```diff
@@ -47,6 +47,14 @@ public VCImplication getNext() {
         return next;
     }
 
+    public boolean hasNext() {
+        return next != null;
+    }
+
+    public boolean hasBinder() {
+        return name != null && type != null;
```
CatarinaGamboa, line 75: so dont all have binders? i think every VCImplication had a name and type, the question might be if they appear inside the refinement, but i didnt think this would be a possibility

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationUtils.java
```diff
@@ -0,0 +1,37 @@
...
+
+import liquidjava.processor.VCImplication;
+import liquidjava.rj_language.Predicate;
+import liquidjava.rj_language.SimplifiedPredicate;
+import liquidjava.rj_language.SimplifiedPredicate.Binder;
+import liquidjava.rj_language.ast.Expression;
+
+class VCSimplificationUtils {
+
+    public static Expression activeExpression(Predicate refinement) {
```
CatarinaGamboa: whats an activeExpression? maybe we can add a method to Predicate to get the expression clone instead of this utils

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationUtils.java
```diff
@@ -0,0 +1,37 @@
...
+        if (refinement instanceof SimplifiedPredicate simplified)
+            return simplified.getSimplifiedPredicate().getExpression().clone();
+        return refinement.getExpression().clone();
+    }
+
+    public static Predicate originPredicate(Predicate refinement) {
```
CatarinaGamboa: Same here

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationUtils.java
```diff
@@ -0,0 +1,37 @@
...
+        return refinement.clone();
+    }
+
+    public static VCImplication copyWithRefinement(VCImplication implication, Predicate refinement) {
+        if (implication.hasBinder())
+            return new VCImplication(implication.getName(), implication.getType(), refinement);
+        return new VCImplication(refinement);
+    }
+
+    public static boolean sameVc(VCImplication left, VCImplication right) {
```
CatarinaGamboa: Equals in VCImplication?

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationUtils.java
```diff
@@ -0,0 +1,37 @@
...
+        return refinement.getExpression().clone();
+    }
+
+    public static Predicate originPredicate(Predicate refinement) {
+        if (refinement instanceof SimplifiedPredicate simplified)
+            return simplified.getOrigin().clone();
+        return refinement.clone();
+    }
+
+    public static VCImplication copyWithRefinement(VCImplication implication, Predicate refinement) {
```
CatarinaGamboa: this could actually be a new constructor of VCImplication right?

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSubstitution.java
```diff
@@ -0,0 +1,139 @@
...
+    private static VCImplication substitute(VCImplication implication, VCImplication source, Expression value) {
+        if (implication == null)
+            return null;
+
+        // skip the source node to remove it from the chain and start substitution from the next node
+        if (implication == source)
+            return substitute(implication.getNext(), source, value);
+
+        Predicate refinement = substituteRefinement(implication.getRefinement(), source, value);
+        VCImplication result = VCSimplificationUtils.copyWithRefinement(implication, refinement);
```
CatarinaGamboa: shouldnt we also set the origin here?

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSubstitution.java
```diff
@@ -0,0 +1,139 @@
...
+        if (isVar(right, name) && !containsVariable(left, name))
+            return Optional.of(new Substitution(implication, left.clone()));
+
+        return Optional.empty();
+    }
+
+    /**
+     * Checks whether an expression is a variable with a given name
+     */
+    public static boolean isVar(Expression expression, String name) {
```
CatarinaGamboa, line 112: public?

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSubstitution.java
```diff
@@ -0,0 +1,139 @@
...
+        return expression instanceof Var var && name.equals(var.getName());
+    }
+
+    /**
+     * Checks whether an expression contains a variable name
+     */
+    public static boolean containsVariable(Expression expression, String name) {
```
CatarinaGamboa: same

liquidjava-verifier/src/test/java/liquidjava/rj_language/opt/VCImplicationGenerator.java
```diff
@@ -0,0 +1,89 @@
...
+import com.pholser.junit.quickcheck.generator.Generator;
+import com.pholser.junit.quickcheck.random.SourceOfRandomness;
+import liquidjava.processor.VCImplication;
+
+public class VCImplicationGenerator extends Generator<VCImplication> {
+
+    static final String[] BINDERS = { "x", "y", "z" };
+    static final String[] FREE_VARS = { "a", "b", "c" };
+    private static final String[] COMPARISON_OPS = { "==", "!=", ">=", ">", "<=", "<" };
+
```
CatarinaGamboa, line 15: then you'll add other operations when you implement those substitutions?
like `&&`, ite, etc?

CatarinaGamboa (changes requested):
Tests look good, some code organization changes

liquidjava-verifier/src/main/java/liquidjava/processor/VCImplication.java
rcosta358: Not all of them. We have this constructor `VCImplication(Predicate ref)` which does not set the `name` and `type`. For example in `∀x:int. x == 1 => x > 0` the first vc has a binder but the second doesnt.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationUtils.java
rcosta358: The active expression is the current expression we are simplifying. Yes we can remove this method and simply clone the current expression.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationUtils.java
rcosta358: Yep, that's better. I'll add it.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSubstitution.java
rcosta358: Yeah simply because we need it in the tests and I didn't want to repeat the same code.

liquidjava-verifier/src/test/java/liquidjava/rj_language/opt/VCImplicationGenerator.java
rcosta358: Yep!

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSubstitution.java
rcosta358: The origin is already set in `substituteRefinement`.

CatarinaGamboa (approved):
Cool! I think this abstraction works well!

PR #250, opened 2026-06-11
Add VC Folding Simplification

rcosta358:
## Description
This PR adds VC simplification for foldable expressions through `VCImplication` chains. When a refinement contains constant arithmetic, boolean expressions, conditionals, adjacent integer constants, or resolved enum literals, it folds the VC node while preserving its origin and the rest of the VC chain.

It also adds focused unit tests for both the VC folding alone and the VC simplification with both the substitution and folding.

## Example
```java
1 + 2 > 0

is simplified to

3 > 0

with origin 1 + 2 > 0

and then simplified to

true

with origin 3 > 0
```

## Related Issue
None.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added/updated tests


liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplification.java
```diff
@@ -36,7 +36,6 @@ public static VCImplication simplifyOnce(VCImplication implication) {
         if (!implication.equals(substituted))
             return substituted;
 
-        // TODO: add more simplification steps here (e.g., folding)
-        return substituted;
+        return VCFolding.apply(implication);
```
CatarinaGamboa, line 39: It could be cool to have like `implication.fold()` and have that method call the `VCFolding`. In terms of API thats a bit more ergonomic since the user doenst need to know that there is another class resposible for the folding itself

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFolding.java
```diff
@@ -0,0 +1,271 @@
...
+
+    /**
+     * Applies folding to the first foldable predicate in a VC chain
+     */
+    public static VCImplication apply(VCImplication implication) {
+        if (implication == null)
+            return null;
+
+        Expression expression = implication.getRefinement().getExpression();
+        Expression folded = fold(expression);
```
CatarinaGamboa, line 29: So with this architecture, each VCImplication only has one origin->simplification right?
Its not possible for one VCImplication to formulate 2 "independent" simplifications for different parts of the expression like you had planned before.
Like
```
           VC: forall x: x == 3 -> forall y: y == 4 -> x < y
            // there are 2 independent simplifcations possible
VC: forall y: y == 4 -> 3 < y      e      forall x: x == 3 -> x < 4
           // but just one outcome for each of the next ones
                             3 < 4
                              true
```
If you want to keep this structure, I think this architecture does not work 😕 
Nevertheless, this current option works for me if you want to leave that for a followup but it might require some more deep changes

CatarinaGamboa (changes requested):

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFolding.java
rcosta358: Yes, you are correct, but I think we should keep this approach because it stays consistent with the `VCSubstitution`, and even though it can be a bit more verbose than the previous one, each simplification step is atomic and easier to debug.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplification.java
rcosta358: Maybe just `implication.simplify()` to call `VCSimplification.simplifyToFixedPoint`?

liquidjava-verifier/src/main/java/liquidjava/processor/VCImplication.java
```diff
@@ -85,6 +86,10 @@ public String toString() {
             return String.format("%-20s %s", "", refinement.toString());
     }
 
+    public VCImplication simplify() {
```
CatarinaGamboa, line 89: 👍

liquidjava-verifier/src/test/java/liquidjava/rj_language/opt/VCFoldingTest.java
```diff
@@ -0,0 +1,181 @@
...
+        assertSimplificationSteps(VCFolding::apply, vc("4 % 0 == 0"), chain(expect("4 % 0 == 0", "4 % 0 == 0")));
+    }
+
+    @Test
+    void leavesRealDivisionAndModuloByZeroUnchanged() {
+        assertSimplificationSteps(VCFolding::apply, vc("4.0 / 0.0 == 0.0"),
+                chain(expect("4.0 / 0.0 == 0.0", "4.0 / 0.0 == 0.0")));
+        assertSimplificationSteps(VCFolding::apply, vc("4.0 % 0.0 == 0.0"),
+                chain(expect("4.0 % 0.0 == 0.0", "4.0 % 0.0 == 0.0")));
+    }
```
CatarinaGamboa, line 56: Maybe more tests for division and module?
I think there might be some issues with things like:
  - `(2 - 7) / 2` 
  - `-5 % 2` 
 since the SMT solver and the Java solving might diverge

CatarinaGamboa (changes requested):
Just add a couple tests for division and %

liquidjava-verifier/src/test/java/liquidjava/rj_language/opt/VCFoldingTest.java
rcosta358: Yes, that's why the `/` and `%` operators are not included in the operators to generate random VCs in the property based tests. The unit tests are fine because they don't actually use the SMT solver to check equivalence between original and simplified expressions.

CatarinaGamboa (approved):
lgtm - discussed offline the design decisions

PR #254, opened 2026-06-14
Refactor VC Simplification

rcosta358:
## Description
This PR refactors VC simplification passes from static to instance classes using a shared `VCSimplificationPass` contract. It also adds `VCExpressionSimplificationPass<C>` as an abstract base for expression simplifications, where `C` is the context carried while simplifying the VC (only used in the `VCArithmeticSimplification` for now) . This removes duplicated VC chain traversal logic from folding, arithmetic, and logical simplification. `VCSubstitution` implements the pass interface directly because it rewrites the whole VC chain rather than simplifying a refinement expression.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Code refactoring

## Checklist
- [x] Added/updated tests


liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCExpressionSimplificationPass.java
```diff
@@ -0,0 +1,48 @@
...
+
+import liquidjava.processor.SimplifiedVCImplication;
+import liquidjava.processor.VCImplication;
+import liquidjava.rj_language.Predicate;
+import liquidjava.rj_language.ast.Expression;
+
+/**
+ * Base implementation for passes that simplify one refinement expression at a time.
+ */
+abstract class VCExpressionSimplificationPass<C> implements VCSimplificationPass {
```
CatarinaGamboa, line 11: Why do we need the generic?  Can you clear it in the class documentation?

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCExpressionSimplificationPass.java
```diff
@@ -0,0 +1,48 @@
...
+
+    protected C initialContext() {
+        return null;
+    }
+
+    protected C nextContext(C context, VCImplication implication) {
+        return context;
+    }
+
+    protected abstract Expression simplify(Expression expression, C context);
```
CatarinaGamboa, line 26: Context? This is kinda confusing since this is not the context we use in the codebase right

CatarinaGamboa (approved):
Comments below, especially the need for the generic and its use as "context" but you can change it in later versions

rcosta358:
The "context" is basically just something we need to keep track during the simplification, not to be confused with the LiquidJava context. I can change it to a different name to avoid the confusion. It needs to be a generic because it can be anything. For example in the `VCArithmeticSimplification` the context is `List<Expression>`, since we need to track the non-zero expressions to avoid performing division and modulo by zero.

CatarinaGamboa:
> The "context" is basically just something we need to keep track during the simplification, not to be confused with the LiquidJava context. I can change it to a different name to avoid the confusion. It needs to be a generic because it can be anything. For example in the `VCArithmeticSimplification` the context is `List<Expression>`, since we need to track the non-zero expressions to avoid performing division and modulo by zero.

So its like what you need to know to perform the transformation?

rcosta358:
Yes.

CatarinaGamboa:
Yeah maybe in a followup explain better in the class documentation and change the variable to something else, maybe `simplificationInput` or anything else to not be confusing

PR #256, opened 2026-06-15
Return Null Origin for Unsimplified VCs

rcosta358:
This PR makes plain `VCImplication` nodes have no origin unless they were actually simplified, instead of an origin equal to the itself. This reduces unnecessary memory usage and will make it easier to implement the simplification expansion in the IDE later. This also simplifies the tests a bit.


liquidjava-verifier/src/main/java/liquidjava/processor/VCImplication.java
```diff
@@ -33,11 +33,14 @@ public VCImplication(VCImplication implication, Predicate ref) {
     }
 
     public VCImplication getOrigin() {
```
CatarinaGamboa, line 35: Counldnt we add these methods just to the SimplifiedVC? instead of returning null?

CatarinaGamboa (approved):

liquidjava-verifier/src/main/java/liquidjava/processor/VCImplication.java
rcosta358: I think we could, but then for each `VCImplication` we would need to check if it was a `SimplifiedVCImplication` and then cast it into before calling the method.

rcosta358: Also since we changed the simplification approach in #262 which removes the `SimplifiedVCImplication`, that doesn't really matter anymore.

PR #259, opened 2026-06-17
Fix VC Substitution

rcosta358:
## Description
This PR fixes `VCSubstitution`, where it treating a binder equality as a substitution even when the binder variable was never used later in the VC chain. This meant that a binder like `∀#ret_8:int. #ret_8 == x + 1` could be incorrectly be removed and turn into `true`. 

## Example
```
∀#ret_8:int. #ret_8 == x + 1

incorrectly simplified to 

true
```

## Related Issue
None.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist
- [x] Added/updated tests


CatarinaGamboa (approved):
lgtm

PR #261, opened 2026-06-17
Simplification Migration

rcosta358:
This PR migrates the old expression-based simplification to the new VC-based simplification.


CatarinaGamboa (approved):
nicee

PR #262, opened 2026-06-18
Track Simplification Origins by Full VC Implication

rcosta358:
## Description

This PR refactors VC simplification origin tracking. Instead of storing origins on individual `VCImplication` nodes, each simplification step now records the complete VC implication chain through a new `VCSimplificationResult`.

## Example

```java
VCSimplificationResult result = VCSimplification.simplifyToFixedPoint(implication);

VCImplication simplified = result.getImplication();
VCSimplificationResult origin = result.getOrigin();
```

## Related Issue

None.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [x] Code refactoring

## Checklist

- [x] Added/updated tests


liquidjava-verifier/src/main/java/liquidjava/processor/VCImplication.java
```diff
@@ -32,21 +33,6 @@ public VCImplication(VCImplication implication, Predicate ref) {
         this.refinement = ref;
     }
 
-    public VCImplication getOrigin() {
```
CatarinaGamboa, line 35: 👍

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationResult.java
```diff
@@ -0,0 +1,44 @@
+package liquidjava.rj_language.opt;
+
+import java.util.Objects;
+
+import liquidjava.processor.VCImplication;
+
+/**
+ * Result of simplifying VC implication chain
+ */
+public final class VCSimplificationResult {
```
CatarinaGamboa, line 10: Couldnt this extend VCSimplification but just with a new field for origin? Then we could use them interchangeably. This also works and maybe is less confusing but it would be nice to use the result without doing `.getImplication`

CatarinaGamboa (approved):
this is easier to explain than the previous ideas

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationResult.java
rcosta358: Yeah, it's definitely less confusing this way, but I can explore that idea in the future.

PR #265, opened 2026-06-22
Track Simplification Pass Names

rcosta358:
## Description
This PR makes each `VCSimplificationResult` save the name of the simplification that produced it, which is derived automatically through `VCSimplificationPass.getName()`. The VS Code extension can then display this information to make simplification results easier to understand.

## Example
```java
result.getSimplification(); // "Substitution" || "Folding" || "Arithmetic" || ...
```

## Related Issue
None.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Code refactoring

## Checklist

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationResult.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationResult.java
@@ -11,14 +11,18 @@ public final class VCSimplificationResult {
 
     private final VCImplication implication;
     private final VCSimplificationResult origin;
+    private final String simplification;
 
     public VCSimplificationResult(VCImplication implication) {
-        this(implication, null);
+        this.implication = Objects.requireNonNull(implication).clone();
+        this.origin = null;
+        this.simplification = null;
     }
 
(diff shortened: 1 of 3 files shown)


liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationResult.java
```diff
@@ -11,14 +11,18 @@ public final class VCSimplificationResult {
 
     private final VCImplication implication;
     private final VCSimplificationResult origin;
+    private final String simplification;
 
     public VCSimplificationResult(VCImplication implication) {
-        this(implication, null);
+        this.implication = Objects.requireNonNull(implication).clone();
```
CatarinaGamboa: Do we use this "requireNonNull" in several places across the codebase? I was under the impression we didn't use that as a standard

CatarinaGamboa (approved):
Nit

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCSimplificationResult.java
rcosta358: Yes we don't, I'll remove it.

PR #266, opened 2026-06-24
Add VC Simplification Debug Log

rcosta358:
This PR adds the debug logs for the new VC simplification and removes the previous methods for the expression based simplification.


CatarinaGamboa (approved):
Lgtm

PR #267, opened 2026-06-24
Fix VC Function Substitution

rcosta358:
## Description
This PR fixes VC function substitution so function equality premises are removed after they are used to rewrite later premises or the conclusion, similarly to normal VC substitution.

## Example

```java
f(x) == 0
f(y) == f(x) + 1

simplified to

f(x) == 0
f(y) == 0 + 1

and now simplifies to 

f(y) == 0 + 1
```

## Related Issue
None.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Code refactoring

## Checklist
- [x] Added/updated tests

### DIFF
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFunctionSubstitution.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFunctionSubstitution.java
@@ -19,9 +19,10 @@ import liquidjava.rj_language.ast.FunctionInvocation;
 public class VCFunctionSubstitution implements VCSimplificationPass {
 
     /**
-     * A substitution discovered from a function invocation equality
+     * A substitution discovered from a function invocation equality. At {@code sourceNode}, remove
+     * {@code sourceEquality} and in the following nodes replace {@code invocation} with {@code replacement}
      */
-    private record Substitution(VCImplication node, FunctionInvocation invocation, Expression replacement,
+    private record Substitution(VCImplication sourceNode, FunctionInvocation invocation, Expression replacement,
             Expression sourceEquality) {
     }
 
@@ -35,8 +36,8 @@ public class VCFunctionSubstitution implements VCSimplificationPass {
 
         if (substitutionOpt.isPresent()) {
             Substitution substitution = substitutionOpt.get();
-            result = substitute(result, substitution.node(), substitution.invocation(), substitution.replacement(),
-                    substitution.sourceEquality());
+            result = substitute(result, substitution.sourceNode(), substitution.invocation(),
+                    substitution.replacement(), substitution.sourceEquality());
         }
         return result;
     }
--- a/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFunctionSubstitution.java
+++ b/liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFunctionSubstitution.java
@@ -1,7 +1,10 @@
 package liquidjava.rj_language.opt;
 
 import static liquidjava.rj_language.opt.VCSimplificationUtils.copyWithRefinement;
+import static liquidjava.rj_language.opt.VCSimplificationUtils.containsExpression;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Optional;
 
 import liquidjava.processor.VCImplication;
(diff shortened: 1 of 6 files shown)


CatarinaGamboa (approved):
I'm fine if this works. Can you add an issue with this situation so that we know we might still need to fix it down the line?

rcosta358:
Sure.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFunctionSubstitution.java
```diff
@@ -18,7 +21,8 @@ public class VCFunctionSubstitution implements VCSimplificationPass {
     /**
      * A substitution discovered from a function invocation equality
      */
-    private record Substitution(VCImplication node, FunctionInvocation invocation, Expression replacement) {
+    private record Substitution(VCImplication node, FunctionInvocation invocation, Expression replacement,
```
CatarinaGamboa: This could have better name since it's only for the substitution of function invocations from equalities. Think "how should I read this object". Right now is smt like "substitute at the node the invocation by the replacement because of the sourceEquality"  is this the best way to read it? Does it make sense like this? Naming objects, Params, types all have a goal in API design

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFunctionSubstitution.java
```diff
@@ -31,32 +35,52 @@ public VCImplication apply(VCImplication implication) {
...
+        result.setNext(substitute(implication.getNext(), node, invocation, replacement, sourceEquality));
         return result;
     }
 
+    /**
+     * Removes the equality conjunct that supplied the substitution, preserving any sibling conjuncts
+     */
+    private VCImplication removeSourceEquality(VCImplication implication, Expression sourceEquality) {
+        List<Expression> remaining = new ArrayList<>(implication.getRefinement().getExpression().getConjuncts());
+        if (!remaining.remove(sourceEquality))
```
CatarinaGamboa: What's does this if mean?

liquidjava-verifier/src/test/java/liquidjava/rj_language/opt/VCFunctionSubstitutionTest.java
```diff
@@ -41,8 +41,8 @@ void doesNotRewriteEarlierNodesFromLaterEquality() {
     void skipsUsedUpEqualityAndUsesNextAvailableEquality() {
         VCImplication implication = vc("f(x) == 0", "f(y) == f(x) + 1", "f(y) == 1");
 
-        assertSimplificationSteps(substitution::apply, implication, step("f(x) == 0", "f(y) == 0 + 1", "f(y) == 1"),
-                step("f(x) == 0", "f(y) == 0 + 1", "0 + 1 == 1"));
+        assertSimplificationSteps(substitution::apply, implication, step("f(y) == 0 + 1", "f(y) == 1"),
+                step("0 + 1 == 1"));
```
CatarinaGamboa, line 45: Why this second step? Shouldn't it be just 1 == 1 ? Probably not related to this change precisely but this step doesnt make much sense to me

CatarinaGamboa (approved):
Left a couple comments. Also, it feels like we are always using `.clone()` is it actually needed? Also we might consider just making it immutable if we are always cloning it.
The tests seek okay Im stamping anyway

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFunctionSubstitution.java
rcosta358: The naming could be improved, even though it's a private record only within the class.
And yes, that's basically how it should be read. The `sourceEquality` is needed to then remove the equality from the VC.

liquidjava-verifier/src/main/java/liquidjava/rj_language/opt/VCFunctionSubstitution.java
rcosta358: Seeing it now, I think the if itself is not needed, since the `sourceEquality` will always be in the list.

liquidjava-verifier/src/test/java/liquidjava/rj_language/opt/VCFunctionSubstitutionTest.java
rcosta358: Here we are just testing the `VCFunctionSubstitution`, not the `VCFolding`.

PR #268, opened 2026-06-24
Prioritize Ternary Folding

rcosta358:
This PR makes ternary folding happen before simplifying branch expressions to avoid unnecessary simplifications on branches that will be discarded anyway.


CatarinaGamboa (approved):
Lgtm

